PR #2, opened 2018-11-23
Fix annotations for sip.array

The-Compiler:
Otherwise, mypy fails with:

    sip.pyi:38: error: Class sip.array has abstract attributes "__getitem__", "__len__"
    sip.pyi:38: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass

### DIFF
--- a/sip.pyi
+++ b/sip.pyi
@@ -15,7 +15,7 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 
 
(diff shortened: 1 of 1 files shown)


stlehmann:
Used an alias instead and implemented it in https://github.com/stlehmann/PyQt5-stubs/commit/466bc68ad10282dccec20add9f6117b83810660d

PR #3, opened 2018-11-23
Add a simple testsuite

The-Compiler:
This allows to drop Python files inside tests/, which then get executed (to make
sure they're bug-free), and checked with the stubs via mypy. If there's any
output from mypy, the test will fail.

Will fail before #2 is merged, and it'd be great if you could turn on Travis CI for this repository!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+python: 3.6
+
+install:
+  - pip install tox
+
+script:
+  - tox
--- a/tests/qtimer.py
+++ b/tests/qtimer.py
@@ -1,0 +1,9 @@
+"""Test for handling of signals.
+
+In the upstream PyQt stubs, signals are normal methods.
+"""
+
+from PyQt5.QtCore import QTimer
+
+timer = QTimer()
+timer.timeout.connect(lambda: None)
--- a/tests/simple.py
+++ b/tests/simple.py
@@ -1,0 +1,1 @@
+from PyQt5 import QtCore
--- a/tests/test_stubs.py
+++ b/tests/test_stubs.py
@@ -1,0 +1,35 @@
+import os.path
+
+import pytest
+from mypy import api
+
+
+TESTS_DIR = os.path.dirname(__file__)
+
+
+def gen_tests():
+    for filename in os.listdir(TESTS_DIR):
+        if filename.endswith('.py') and not filename.startswith('test_'):
+            yield filename
+
+
+@pytest.mark.parametrize('filename', list(gen_tests()))
+def test_stubs(filename):
+    """Run mypy over example files."""
+    path = os.path.join(TESTS_DIR, filename)
+    stdout, stderr, exitcode = api.run([path])
+    if stdout:
+        print(stdout)
+
+    assert not stdout
+    assert not stderr
+    assert exitcode == 0
+
+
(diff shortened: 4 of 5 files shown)


stlehmann (commented):
Thanks for writing this test-suite. That is definetely something we need. I will gladly implement Travis CI for it. However I have a problem when I try to run the tests with pytest. I get two errors:

    assert not "tests\\qtimer.py:6: error: No library stub file for module 'PyQt5.QtCore'\ntests\\qtimer.py:6: note: (Stub files are from https://github.com/python/typeshed)\n"

and 

    tests\simple.py:1: error: Module 'PyQt5' has no attribute 'QtCore'

With tox it works well but it should also be possible to run the tests with pytest. Do you have a suggestion to this?

The-Compiler:
You'll need to install the package (e.g. via `pip install -e .`) so mypy actually uses the stubs.

stlehmann:
Ah, the classic ... Thanks that worked.

So if I get it right the idea of this testsuite is to add tests for each module and class? For such a big framework that really is a tremendous amount of work. Also there is always the risk of missing things. I think there should be a much more general approach here. It tested the stubs only with recursive package checking via `mypy -p PyQt5-stubs` and it provides very detailed results for the complete package. If I think about it I don't think it is necessary to add unit testing to test if the stub files are correct. Only running mypy over the whole package should actually do the job. Or am I missing something here?

The-Compiler:
> So if I get it right the idea of this testsuite is to add tests for each module and class? For such a big framework that really is a tremendous amount of work. Also there is always the risk of missing things.

The idea was to add some tests for known issues, i.e., fixes done in this repository over the upstream PyQt5 stubs. This is helpful for various reasons:

- It serves as a nice documentation of what changes (or fixes for what issues) there are.
- When updating the stubs from a newer PyQt/Qt version, it provides at least some guarantees that there are no regressions for things already fixed before.

So, the idea isn't to test everything possible (only some samples) - it never is, tests can only show the presence of bugs, not the absence :wink:

> It tested the stubs only with recursive package checking via `mypy -p PyQt5-stubs` and it provides very detailed results for the complete package.

Ah, I wasn't aware that was possible! That'd probably be useful as an additional step to add to Travis (but it shows quite some errors right now). Still, it only shows that the stubs are consistent/syntactically valid in themselves - doesn't mean they're necessarily correct.

> If I think about it I don't think it is necessary to add unit testing to test if the stub files are correct.

Depends on what you mean with "correct". Right now, they run without syntax errors, but they still have various issues. I have [fixes for what I found](https://github.com/qutebrowser/PyQt5-stubs/commits/wip) and I'm also trying to upstream issues which are not mypy-specific ([1](https://www.riverbankcomputing.com/pipermail/pyqt/2018-November/041139.html), [2](https://www.riverbankcomputing.com/pipermail/pyqt/2018-November/041144.html), [3](https://www.riverbankcomputing.com/pipermail/pyqt/2018-November/041145.html), [4](https://www.riverbankcomputing.com/pipermail/pyqt/2018-November/041146.html)).

At some point I'd like to update the stubs to PyQt/Qt 5.12, but doing so without having at least some reassurance that things won't break again is much harder.

stlehmann:
I added the testsuite to the branch pyqt5.11.3

stlehmann:
Thanks :)

Issue #4, opened 2019-01-20
QCoreApplication aboutToQuit not defined as signal


karlch:
First of all thanks a bunch for your work on this!

When connecting to the `aboutToQuit` signal in `QCoreApplication` mypy complains with:
```
"Callable[[], None]" has no attribute "connect"
```

I suspect this is because it is simply defined as a function in the stub, not as a signal. Is there any reason for this or could this be changed?

stlehmann:
Thanks for the feedback. The upstream PyQt5 stubs type-annotate signals as functions. They need to be changed to the type *pyqtSignal*. I will change this right away. If you discover any more such issues feel free to make a PR or report.

stlehmann:
Fixed with 9e7cbb938ba097a55cbb7aaf8de23170a54dce1d

karlch:
Perfect, thanks!

3rdcycle:
> Fixed with [9e7cbb9](https://github.com/stlehmann/PyQt5-stubs/commit/9e7cbb938ba097a55cbb7aaf8de23170a54dce1d)

Wait, am I missing something here? This provided a fix for `applicationVersion` but none of the other signals including the one originally requested (`aboutToQuit`)? As far as I can see most signal stubs are still simple `def`s.

stlehmann:
You are right. There went something wrong, here. I'll have a look at it.

3rdcycle:
That was quick 😃 thanks.

PR #7, opened 2019-05-07
Define signals in QtPrintSupport as pyqtSignal

karlch:
Similar to issue #4 the signals in this module were defined as functions by the upstream PyQt5 stubs. This commit fixes this by re-defining them as signals. In addition the signals in QDialog were updated as it is used as parent in the print dialog classes.

Not sure if there is anything else that needs to be done other than simply redefining the methods. I am especially unsure because of the overloaded methods which I changed into simple signals. If so, just let me know!


stlehmann (commented):
Looks alright to me.

PR #8, opened 2019-07-29
Some signals for QtCore, QtWidgets, QTNetworking added QtSerialPort newly created

Schleehauf:
In the original version the signals were contained as methods.


stlehmann:
Your PR seems to be broken. Please make sure the diff contains only the lines that you changed. Otherwise it is impossible for me to tell which lines you changed. Also please make sure that the checks are passed. Thx.

Schleehauf:
Please carefully review the new QtSerialPort.pyi, I am not sure if I fully understood how to put in Enums and flags.

stlehmann (approved):

stlehmann:
Please do not delete the original lines. Instead comment them out and add a note which issue you addressed by changing them.

stlehmann (changes requested):
Please comment out the original code lines instead of deleting them. Also add a comment which issue (see issues.md) you are addressing. Have a look at the recent code changes for an example.

PR #13, opened 2019-10-23
Fix signature of pyqtSlot

The-Compiler:
The `@pyqtSlot` decorator accepts arguments. In other words, pyqtSlot is:

- A callable (taking types and other arguments)
- Returning the "real" decorator (another callable)
- Which then takes a callable (the decorated function) and returns
  it (unmodified)

In 401a0f19353e82754f17a3d4f86f2714057fa278, its signature was changed.

However, that fix was incorrect, as it claimed that `@pyqtSlot(...)` replaces
the decorated function with an `Optional[str]`. This causes code like:

    @pyqtSlot(str)
    def func(s: str) -> None:
        pass

    func("test")

To fail with:

    error: "str" not callable
    error: "None" not callable

which is incorrect.

The original `-> typing.Callable[..., typing.Callable[..., None]]` annotation
was more correct, but it has two problems:

- It restricts slots to functions with a `None` return value. That however is
  not correct, as slots can still be used as normal functions.
- It didn't preserve the signature of the decorated functions, leading to weaker
  typechecking.

This commit changes the annotation to one which:

- Does the correct thing for `@pyqtSlot`
- Allows any callable (via `bound=typing.Callable`)
- Preserves the signature of the decorated function (via `TypeVar`)


The-Compiler:
Tests are likely to fail, but that's for unrelated reasons.

stlehmann:
Thanks for the PR :)

PR #18, opened 2019-11-05
Add support for signals

Djedouas:
All pyqt signals are now typed as such


Djedouas:
I ran into a problem when connecting QMenu().aboutToShow() signal :
`projectMenu.aboutToShow.connect(self.controller.saveSettings)`

Which creates a mypy error:
` news_view.py:93:9: error: "Callable[[], None]" has no attribute "connect"`

So I decided to patch all others signals in `QtWidgets.pyi`, like it was already done for `QAbstractButton`

The-Compiler:
Did you automate this in some way? Why only `QtWidgets.pyi`?

Djedouas:
It's handmade. That's why it's only `QtWidgets.pyi`.
Therefore it can not be considered completely exhaustive, like before.

stlehmann:
Thanks for your effort of manually adding the signals. I know that's a tedious work. I would like to have a way to automate this stuff. Thanks again :)

Djedouas:
Thanks a lot!
Happy to participate ;-)

Issue #6, opened 2019-04-01
Automating patching/generating of stubs


zdenop:
Can you please  create new version for recent version (5.12.1) PyQt5? 
Your package requires PyQt5==5.11.3

stlehmann:
Done :)

stlehmann:
Some sort of automation would be great here.

zdenop:
Thanks. IMO also https://github.com/stlehmann/PyQt5-stubs/tree/5.12.1.0/requirements should be updated....

kazum1kun:
Qt 5.13 is out, would you consider updating it again to latest Qt? Thanks!

bjwest:
Seriously!  Should we need to ask for this?  I understand you may be busy so how about automating the process somehow?

The-Compiler:
@bjwest Your tone seems unnecessarily demanding for something someone does for free, in their spare time. Everyone (including @stlehmann, see the comment above) seems to agree that more automation seems like a good idea, so why not start contributing to make the process easier (and thus faster for everyone)?

stlehmann:
@bjwest You are very welcome to present some sort of automation concept here. 
@The-Compiler thanks for pointing that out.

I would very much like to automate the process but it would be good to have someone with experience in this matter. I think the first step would be to create a hook to the PyQt5 repository that creates a new branch and merges the current PyQt5 stubs. Do you guys know someone with experience in git hooks?

The-Compiler:
@stlehmann IMHO the best way forward would be to see what kind of stub changes makes sense to contribute upstream to sip (maybe hidden behind a commandline flag, if there's a reason to diverge between IDE-friendly autocompletion stubs and mypy-friendly type checking stubs).

Are you at Europython? If so, let's talk or maybe even work on PyQt stubs at the sprints?

(Also, note that there isn't a public PyQt5 repository, you'd need to get new releases from the website or PyPI)

stlehmann:
@The-Compiler I would really love to got to Europython but sadly I couldn' t make it :(

What you're suggesting means we should initiate some communication with some PyQt5 guys. That might actually be a good way to go. This way we tackle the issues at their roots. Has somebody got some good contacts to PyQt5 development team?

The-Compiler:
The [PyQt mailing list](https://www.riverbankcomputing.com/mailman/listinfo/pyqt) is the best way to get in contact with the author - I've done so before and provided some observations (before you started PyQt5-stubs IIRC).

[...]

PR #19, opened 2019-12-04
Widen PyQt5 dependency to 5.13.*

welps:
This doesn't quite address #6 but I think we should be able to agree that the PyQt5 dependency version can be widened to patch/maintenance releases?


sweethuman (approved):

welps:
@stlehmann Can you merge when you get a chance?

stlehmann:
Yes I think we can do that. Thanks for the PR.

PR #22, opened 2019-12-18
Define signals correctly in QProcess

bluebird75:
signals in QProcess were declared as funcitons. This patch fixes by declaring them as signals.


stlehmann (approved):

stlehmann:
Thanks for the PR ;)

PR #26, opened 2020-02-20
Add operator support for QPoint and QPointF. 

KmolYuan:


PyQt5-stubs/QtCore.pyi
```diff
@@ -5994,6 +5994,10 @@ class QPoint(sip.simplewrapper):
 
     def __neg__(self) -> 'QPoint': ...
     def __pos__(self) -> 'QPoint': ...
+    def __add__(self, point: 'QPoint') -> 'QPoint': ...
+    def __sub__(self, point: 'QPoint') -> 'QPoint': ...
+    def __mul__(self, point: 'QPoint') -> 'QPoint': ...
```
The-Compiler: Multiplying a point by a point doesn't make any sense:

```
>>> QPoint(1, 2) * QPoint(1, 2)
TypeError: unsupported operand type(s) for *: 'QPoint' and 'QPoint'
```

This should probably be `factor: float` or so?

PyQt5-stubs/QtCore.pyi
```diff
@@ -5994,6 +5994,10 @@ class QPoint(sip.simplewrapper):
...
+    def __truediv__(self, point: 'QPoint') -> 'QPoint': ...
```
The-Compiler: Why `__truediv__` rather than `__div__`? Also, same as above.

PyQt5-stubs/QtCore.pyi
```diff
@@ -6019,6 +6023,10 @@ class QPointF(sip.simplewrapper):
 
     def __neg__(self) -> 'QPointF': ...
     def __pos__(self) -> 'QPointF': ...
+    def __add__(self, point: 'QPointF') -> 'QPointF': ...
+    def __sub__(self, point: 'QPointF') -> 'QPointF': ...
+    def __mul__(self, point: 'QPointF') -> 'QPointF': ...
+    def __truediv__(self, point: 'QPointF') -> 'QPointF': ...
```
The-Compiler: same as above for `__mul__` and `__truediv__`.

The-Compiler (changes requested):

PyQt5-stubs/QtCore.pyi
KmolYuan: Sorry, it a typo. Float is right.

PyQt5-stubs/QtCore.pyi
KmolYuan: I think use Python 3 operator (`__truediv__` and `__floordiv__`) will be better than Python 2 (`__truediv__` and `__div__`).

The-Compiler: Ah, true, I read the docs wrong there. My bad!

The-Compiler (approved):

stlehmann (approved):

Issue #1, opened 2018-11-21
Dealing with upstream PyQt upgrades


The-Compiler:
Hey! I wanted to start a project like this [for a while](https://www.riverbankcomputing.com/pipermail/pyqt/2016-September/038070.html) (whoa, it's been two years already?!) but never got around to it. Now, as part of an [university project](https://lists.schokokeks.org/pipermail/qutebrowser-announce/2018-September/000051.html) I wanted to introduce mypy to [qutebrowser](https://qutebrowser.org/) and found this project.

This seems like a great start at making those stubs actually usable with mypy - thanks a lot for that!

I'll probably start contributing some pull requests over the next couple of days, for some issues I've seen when running mypy over qutebrowser.

I'm also wondering: What version of PyQt are those stubs based on? How do you intend to upgrade to newer versions? Maybe it'd be worthwhile to introduce an `upstream` branch with PyQt's files commited 1:1 for every PyQt release, and then those commits with the PyQt update can be cherry-picked to `master` in order to update the modified bindings?

stlehmann:
Thanks for contacting me. I use PyQt5 a lot for my GUI projects and wanted to benefit from static type-checking. That's why I created this repository in the first place.

You can find the version of PyQt that the stubs are based on in the setup.py as defined by [PEP561](https://www.python.org/dev/peps/pep-0561/#stub-only-packages).

Of course it would be good to keep track of upstream changes but honestly I couldn' t bring up a good idea how to manage that efficiently. The thing is that the stub files returned when compiling PyQt5 are very buggy and there are a lot of changes to them that need to be made before they are actually usable.

The-Compiler:
setup.py claims that it's based on PyQt 5.11.3 stubs, but that doesn't seem to be the case - things like enum numbers added in Qt 5.8 (say, `AA_DontCheckOpenGLContextThreadAffinity` in [Qt::ApplicationAttribute](http://doc.qt.io/qt-5/qt.html#ApplicationAttribute-enum)) aren't in the stubs, and the copyright also says 2016. It seems to me like they were generated around PyQt 5.7, not 5.11.

stlehmann:
Mh, that might be the case...Honestly, I didn't watch the version so closely. But you' re right this needs to be taken care of in the future.

stlehmann:
I updated all stubs to 5.11.3.

stlehmann:
The original upstream stubs can be found on branch 5.11.3.org

[...]

PR #29, opened 2020-03-15
Update to PyQt 5.14.1

BryceBeagle:
I updated the stubs, merging in the changes between the generated 5.13.1 and 5.14.1 stub files. 

Please let me know if there's anything else I can do. Should I also publish the upstream stubs to the upstream branch?


BryceBeagle:
@The-Compiler @stlehmann 

I'm not sure who's the current maintainer for this project but I'd like to get this pull request reviewed. Is there anything that I need to do?

PyQt5-stubs/QtCore.pyi
```diff
@@ -1650,6 +1650,10 @@ class QObject(sip.wrapper):
 
     def __init__(self, parent: typing.Optional['QObject'] = ...) -> None: ...
 
+    @typing.overload  # type: ignore  # fixes issue #1
```
The-Compiler, line 1653: I don't understand the comment here - what issue #1?

tox.ini
```diff
@@ -5,6 +5,7 @@ envlist = py36
 deps =
   mypy
   pytest
+  pip~=20.0
```
The-Compiler: This probably should be `>=`, no?

The-Compiler (commented):
I still have no idea how the current workflow in this repo looks like (and haven't had time to actually do something myself so far) - but here's a quick review. LGTM otherwise!

PyQt5-stubs/QtCore.pyi
BryceBeagle: Yeah that's a little weird isn't it? It's referring to the first issue in [here](https://github.com/stlehmann/PyQt5-stubs/blob/master/issues.md). I was just following the precedent that already existed in the repo.

Some examples: [1](https://github.com/stlehmann/PyQt5-stubs/blob/bd0fe7ce030e9c97fe046878351b34a8a7ec8de6/PyQt5-stubs/QtPrintSupport.pyi#L118), [2](https://github.com/stlehmann/PyQt5-stubs/blob/bd0fe7ce030e9c97fe046878351b34a8a7ec8de6/PyQt5-stubs/QtSql.pyi#L484), [3](https://github.com/stlehmann/PyQt5-stubs/blob/cd4148a3ca213a895ddbc3431247088dad047ae0/PyQt5-stubs/QtCore.pyi#L1941)

I can remove it if you think it's unnecessary

tox.ini
BryceBeagle: I usually like to lock my package requirements to a major release with `~=`. If you think we'd be ok with any future pip release I can change it to `>=`, as suggested.

PyQt5-stubs/QtCore.pyi
stlehmann: The comment refers to the collection of common issues in https://github.com/stlehmann/PyQt5-stubs/blob/master/issues.md and adresses overload errors. Issue #1 adresses overload decorators. I wanted to keep track of common issues so they could be fixed upstream someday.

tox.ini
stlehmann: Why do you add the pip-dependency at all? CI works fine as it is and pip ist installed on default anyway

BryceBeagle: If  you look at the commits for this pull request, CI didn't work until I added the pip dependency to tox. I think this is because the latest versions of PyQt5 are packaged in such a way that old versions of pip are unable to install them.

BryceBeagle: It wasn't that I was _adding_ pip as a dependency -- I was setting it to use a later version than tox was using by default. Maybe there's a better way to tell tox what version of pip to use?

stlehmann: @BryceBeagle I didn't see that the CI was broken before. So they must have changed something in the setup of PyQt5 with version 5.14. Anyway I would go for PyQt5 >= 20.0 to be future-proof. If we run into any issues with future CI-version it's better to know from CI than from users.

stlehmann:
Thanks for your PR and sorry for the delay.

BryceBeagle:
No problem. Should I also make a PR to bump the upstream branch to 5.14.1?

stlehmann:
Yes this would be great 👍

tox.ini
The-Compiler: What changed is that PyQt started to use a [manylinux2014](https://www.python.org/dev/peps/pep-0599/) wheel tag (the previous manylinux1 tag was always inaccurate). However, support for that was only added in [pip 19.3](https://github.com/pypa/pip/blob/master/NEWS.rst#193-2019-10-14).

PyQt5-stubs/QtCore.pyi
The-Compiler: Ah, right. I definitely support having such an issues.md (IIRC I even originally suggested that). I thought of GitHub issues when reading "Issue #1", which didn't make much sense. Probably fine to leave it as-is though.

PR #33, opened 2020-04-08
Added equality operators to QTableWidgetItem

TilmanK:


stlehmann (changes requested):
mypy check fails because of incompatible supertype:

```
py37 run-test: commands[1] | mypy -p PyQt5-stubs

PyQt5-stubs/QtWidgets.pyi:8823: error: Argument 1 of "__eq__" is incompatible with supertype "object"; supertype 

defines the argument type as "object"

PyQt5-stubs/QtWidgets.pyi:8823: note: It is recommended for "__eq__" to work with arbitrary objects, for example:

PyQt5-stubs/QtWidgets.pyi:8823: note:     def __eq__(self, other: object) -> bool:

PyQt5-stubs/QtWidgets.pyi:8823: note:         if not isinstance(other, QTableWidgetItem):

PyQt5-stubs/QtWidgets.pyi:8823: note:             return NotImplemented

PyQt5-stubs/QtWidgets.pyi:8823: note:         return <logic to compare two QTableWidgetItem instances>

PyQt5-stubs/QtWidgets.pyi:8828: error: Argument 1 of "__ne__" is incompatible with supertype "object"; supertype defines the argument type as "object"

Found 2 errors in 1 file (checked 12 source files)
```
If I get it right it is recommended to use object as argument type for equality operators. So please change the type to object and this should work.

stlehmann:
Looks fine for me, now. Thanks for the PR :)

TilmanK:
Sure. Thanks for the library ;)

PR #36, opened 2020-04-24
Dockerfile to build upstream stubs

BryceBeagle:
Here's a draft pull request for a Dockerfile to build the upstream stubs, along with the new/updated stubs. Some existing stubs receive new annotations to work data types from the new stubs.

There's also a build script (`build_upstream.py`) that handles building the image and extracting the files to the host.

Some things worth discussion:
* No idea how to build only stubs with the new `sip-install` command. The stubs are all generated at the beginning of the PyQt build and then I have to wait for it to finish. I'm sure there's a way I could kill the build early, but I didn't bother.
* I'm unable to build `QtAxContainer`, `QtAndroidExtras`, `QtMacExtras`, and `QtWindowsExtras` in Docker, to my understanding
* Even though I install their dependencies, I'm missing stubs for the following. Maybe they don't generate?
    * `Qt3d`
    * `QtCharts`
    * `QtDataVisualization`
    * `QtPurchasing`
    * `QtWebEngine`
    * `Enginio` (deprecated so probably doesn't matter)

For reference: all the modules that PyQt supports:
https://www.riverbankcomputing.com/static/Docs/PyQt5/module_index.html#ref-module-index


The-Compiler:
PyQtWebEngine is its [own package](https://pypi.org/project/PyQtWebEngine/), so to get QtWebEngine stubs, you'd also need to download/build that.

stlehmann:
That looks great to me. I didn't know of this sip-install command. That added a bunch of new PyQt5-stubs which is great.

stlehmann:
@BryceBeagle Would you like to add QtWebEngine to the build?

BryceBeagle:
Ah, I didn't realize that there were other PyQt packages. I looked at it briefly, but it looks like `PyQtWebEngine`'s sources are only available for 5.14.0. I'd assume they are compatible with 5.14.2?

Also, because it's a different base package, wouldn't the stubs for them need to go in a package named `PyQtWebEngine-stubs`?

stlehmann:
> Ah, I didn't realize that there were other PyQt packages.

I didn't know that this Python package existed either.

From my point of view I think it makes sense to include it in our PyQt5-stubs if it lives in the same namespace (PyQt5.QtWebEngine) and if we can include the stub generation in our build process. I take it one always needs to install PyQt5 along with PyQtWebEngine so it would be more convenient to keep all stubs together in one place.

BryceBeagle:
Added: 
- Qt3DAnimation.pyi
- Qt3DCore.pyi
- Qt3DExtras.pyi
- Qt3DInput.pyi
- Qt3DLogic.pyi
- Qt3DRender.pyi
- QtChart.pyi
- QtDataVisualization.pyi
- QtPurchasing.pyi
- QtWebEngine.pyi
- QtWebEngineCore.pyi
- QtWebEngineWidgets.pyi

The-Compiler:
Sounds good to me as well to include them in the existing package. IIRC the reason they're split upstream is to keep the size of the PyQt5 package small - that's not something we're concerned about with the stubs, and having separate stub packages would probably just make everything more complex for everyone.

BryceBeagle:
Now the only things we're missing are the platform-dependent extras. Maybe that should be a separate pull request?

stlehmann:
I'm a bit lost. Which platform-dependent extras do you refer to?

stlehmann:
Also you could remove the old buildenv directory if you like because the new one is so much better and I think we don't need the old one anymore. 👍

BryceBeagle:
> I'm a bit lost. Which platform-dependent extras do you refer to?

I'm not sure how to build `QtAxContainer`, `QtAndroidExtras`, `QtMacExtras`, and `QtWindowsExtras` within a docker environment

BryceBeagle:
> old buildenv directory

Oh man I didn't even see that this existed because I was only paying attention the `upstream` branch, haha. Are there any aspects/features this has that you would like in the build system I created? 

I should probably add a bit a documentation.....

stlehmann:
> I'm not sure how to build QtAxContainer, QtAndroidExtras, QtMacExtras, and QtWindowsExtras within a docker environment

I wouldn' t bother too much about them. I guess they might only build in their native os.

> Are there any aspects/features this has that you would like in the build system I created?

Actually I think your script is doing very nicely and is even more sophisticated. So it is a great replacement.

stlehmann:
> I should probably add a bit a documentation.....

Yes, some documentation for using the build script would be great. I think this can go right in the README. Also please add a new Changelog entry. Then we should be set up for merging.

BryceBeagle:
> Also please add a new Changelog entry

How do you want to manage this with two branches?

stlehmann:
@BryceBeagle thanks for this PR. That is a great progress because we now have much more modules that we can ship with PyQt5-stubs. Hopefully I was able to merge it the right way. I think it's a good idea to keep the Upstream branch close to master. The only difference should be that the upstream branch contains the raw stubs generated by sip while master contains the altered stubs.

Thanks again :)

PR #41, opened 2020-05-01
Include script for upstream stub generation

stlehmann:
This PR includes the scripts for building upstream stubs from PR #36 in the master branch


stlehmann:
Wow this was a bit messy but hopefully I managed to merge master in upstream by keeping the  generated upstream stubs. We should remember to apply changes directly to master and use upstream only to keep the upstream stubs for comparison.

PR #51, opened 2020-07-08
Add pyqtBoundSignal.signal: str

altendky:


stlehmann:
Could you please add a description to your PR to help others understand why you made the change. Thanks :)

altendky:
Over in QTrio I am comparing bound signals against each other to see if they are 'the same signal'.  While PySide2 already implements this, PyQt5 did not (though it will be in the next release).  I wrote the workaround below which uses the `.signal` attribute which is a string.  Given that I will not be using this attribute soon it could certainly be argued that this PR is not needed anymore.  Or, it could be argued that being more complete is good.  I don't know where the balance lies in this case.

https://www.riverbankcomputing.com/pipermail/pyqt/2020-July/043064.html

https://github.com/altendky/qtrio/blob/03742c3599b27645b81ce51f3f4776888400a259/qtrio/_core.py#L163-L185
```python
    def is_from(self, signal: SignalInstance) -> bool:
        """Check if this emission came from `signal`.
        Args:
            signal: The signal instance to check for being the source.
        """

        # TODO: `repr()` here seems really bad.
        if qtpy.PYQT5:
            return self.signal.signal == signal.signal and repr(self.signal) == repr(
                signal
            )
        elif qtpy.PYSIDE2:
            # TODO: get this to work properly.
            return bool(self.signal == signal)

        raise qtrio.QTrioException()  # pragma: no cover

    def __eq__(self, other):
        if type(other) != type(self):
            return False

        return self.is_from(signal=other.signal) and self.args == other.args
```

The-Compiler:
FWIW I also use `.signal` in the qutebrowser codebase (IIRC to log the name of a signal to the debug log).

stlehmann:
I honestly didn't know about this attribute. Found a reference in here: https://www.riverbankcomputing.com/static/Docs/PyQt5/signals_slots.html. I think we can accept this PR to be more complete.

PR #56, opened 2020-09-12
Improve bound and unbound signal hints

altendky:
`pyqtSignal` is a descriptor and does not have `emit`, `connect`, and `disconnect` methods.  Rather it is used as a class attribute and when you access `an_instance.the_signal` you get a `pyqtBoundSignal` instead of the original `pyqtSignal`.  So the bound signal needs those methods and the unbound signal needs `.__get__()`.

Draft for:
- [x] Maybe we want to get the build passing before making big changes?
- [x] I should look at adding tests related to this.
  - They are pretty minimal, but something.


altendky:
Similar change suggested for PySide2.

https://bugreports.qt.io/browse/PYSIDE-1318?focusedCommentId=527071&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-527071

altendky:
@stlehmann, @BryceBeagle, aside from whether this is correct at the moment or not (still need to add tests etc) but would you rather take signal improvements piecemeal or try to work through a 'complete' solution from the get go?  Basically, work through this or just go all out and come up with a solution for #68?  I would personally default to taking multiple steps since I'm not sure what the whole solution is.  `:]`

BryceBeagle:
I think piecemeal is probably better, so changes can keep rolling out. Functionality is never _lost_ by doing this.

We could make a meta tracking issue if we feel there are many signal-based features to implement.

tests/test_stubs.py
```diff
@@ -31,4 +31,4 @@ def test_files(filename):
     path = os.path.join(TESTS_DIR, filename)
     with open(path, 'r') as f:
         code = f.read()
-    exec(compile(code, filename, 'exec'))
+    exec(compile(code, filename, 'exec'), {})
```
BryceBeagle, line 34: What's the purpose of setting `globals` arg?

BryceBeagle (approved):
Looks good to me. Just a small question that probably won't affect approval

tests/test_stubs.py
altendky: It complained that the name `QtCore` didn't exist without it.  Which is odd.  Worked fine in a REPL as it was but not in pytest.  But mostly I don't understand why things are done this way vs. regular old tests.  I expected at some point to take a pass at revamping the testing but if you can explain the point of this setup with exec now, by all means go for it.

BryceBeagle: Nope I have no clue why it's done this way haha

PR #57, opened 2020-09-12
Slots can return any `object`

altendky:
There is no requirement that slots return specifically `None`, any `object` will do.

### DIFF
--- a/PyQt5-stubs/QtCore.pyi
+++ b/PyQt5-stubs/QtCore.pyi
@@ -42,7 +42,7 @@ class pyqtBoundSignal:
 
 # Convenient type aliases.
 PYQT_SIGNAL = typing.Union[pyqtSignal, pyqtBoundSignal]
-PYQT_SLOT = typing.Union[typing.Callable[..., None], pyqtBoundSignal]
+PYQT_SLOT = typing.Union[typing.Callable[..., object], pyqtBoundSignal]
 
 
 class QtMsgType(int): ...


BryceBeagle:
Can this not be `Callable[..., Any]` or even just plain `Callable`?

altendky:
`Any` means that you can do anything you want with the value which isn't a good default since you don't know what it will be.  `object` means it can be any object and you can only expect to do things to it that you can do with object, which is mostly the case (pending objects that block access to `object` features).

In this case it probably doesn't make much difference, but `object` is a more sensible default as it doesn't effectively disable type hints.  Bare `typing.Callable` just defaults to the `Any`.

The-Compiler:
I'm not really convinced `object` (over `Any`) is the right thing to do here - that being said, I've never used slot return values. But surely you'd usually know what type you get from a specific slot and then do things based on that type? If you get `object` back, like you say, you pretty much can't do anything with the return value, no?

altendky:
`object` forces you, the only person that has any basis for knowing, to explicitly state that you know via a `typing.cast()`.  `Any` leaves you ignorantly passing type hint checks even if you erroneously use the result as if it is two different types.

But does this hint leak back out to anything outside?  My impression was that it just describes what sorts of things are accepted in various places that need a slot.  I think the next layer is to get into `Generic`s so that we can actually relate the type of the incoming callable and properly hint the outputs.  But I haven't even started to think about where or how that would look.  I expect that to be a tough task.

PR #61, opened 2020-09-13
Some cleanup related to mypy errors

altendky:
Draft for:
- [ ] At least let's get through the upstream merges before cleanup.


altendky:
I'm going to handle these as groups of single 'types' of errors instead of mashing them together.

BryceBeagle:
I think splitting up is a good idea. Makes a better paper trail for when we finally get around to communicating with upstream on the issues.

PR #63, opened 2020-09-14
Build 5.15.0

altendky:
Draft for:
- [x] Rebuilding SIP stubs
- [x] Unbreak CI back to its regular brokenness
   - Nevermind, in this case upstream 5.15.0 really just is that broken.


altendky:
Am I supposed to rebuild the `PyQt5-Stubs/sip.pyi` somehow?

BryceBeagle:
I think building the sip stubs needs to be added to the Dockerfile still

altendky:
@BryceBeagle is it documented how to build them, or does that need figured out as well?  I'm just starting to look around a bit.

altendky:
@stlehmann, it looks like you have committed all the updates to `sip.pyi`.  I would have thought it would have come from `PyQt5-sip` but I haven't figured out any way to get it from there.  Perhaps I'm looking in the wrong place?  Once I figure out how to get or make it I'll certainly add it to the `Dockerfile`.

altendky:
Nevermind, looks like it's in SIP proper and included in any install.

altendky:
Do you all get emails when I remove the draft mark?  I'm not sure if I need to ping with a comment like this as well to generate a notification.

BryceBeagle:
I turned off all email notifications, but I don't know if I got an in-Github notification for that specific event either. The notification log just shows that _something_ happnened in a PR, but not _what_.

BryceBeagle (approved):

Issue #40, opened 2020-04-28
PyQt5 wheels will be shipped with stubs in the future


The-Compiler:
According to the PyQt mailinglist, the next PyQt release will ship wheels with its own stubs:

https://www.riverbankcomputing.com/pipermail/pyqt/2020-April/042854.html

My first thought was "hey, cool, so we don't need to build anything to get the stubs files!" (cc @BryceBeagle)... However, my second thought was "wait a minute... what will happen when people have both PyQt5 and PyQt5-stubs installed"? No idea which of the two will take precedence then, even after [reading the relevant part of the PEP](https://www.python.org/dev/peps/pep-0561/#type-checker-module-resolution-order) I don't understand what precedence stubs as part of the installed package (i.e. not inline types) have...

stlehmann:
That will be interesting to see. I'm also curious what quality those stub files will have.

The-Compiler:
> I'm also curious what quality those stub files will have.

I'm guessing they will be the files `sip` already generates.

stlehmann:
> I'm guessing they will be the files sip already generates.

So it will still be necessary to elaborate them for the PyQt5-stubs packages. But it makes no sense if the PyQt5 stubs take precedence over our stubs. This might render this package useless.

BryceBeagle:
I'm pleased that the wheels will begin to ship with stubs! A little disappointing that my build script will get deprecated so quickly though, haha.

stlehmann:
Let's see how it works out with the new stubs. Until now they needed heavy refactoring from our side to be of any use.

altendky:
Having minimally looked over these stubs and the upstream what I have in my head is to get our master fixed up a bit (tests passing?  `:]`) and then talk to Phil about how he might be interested in working through the differences.

altendky:
FYI, shared the work here with Phil.  My feeling is that we'll have to break it down or submit patches, but perhaps I'm misreading.

https://www.riverbankcomputing.com/pipermail/pyqt/2020-October/043291.html

bluebird75:
To answer one of the initial questions, if you install pyqt5-stubs, our stubs take precedence over the stubs shipped directly with PyQt5 . This is a good thing for us, although slithly surprising for me.

Submitting our improvements as patches could certainly be done. But since the other stubs are generated, this will cerainly complicate the generation process for Phil. We shall see when we do it.

bluebird75:
[...]

PR #66, opened 2020-09-14
Get sip.pyi in the Docker build

altendky:


Dockerfile
```diff
@@ -95,6 +100,25 @@ RUN sip-install \
...
+FROM build-dep AS sip
+
+# Reuse arguments from previous build scope
+ARG SIP_VERSION
+ARG SIP_ABI_VERSION
+
+# Install SIP including stubs
+# TODO: Find way to build only stubs. This takes way too long
+WORKDIR /upstream/
+RUN pip install --no-deps --target build sip==${SIP_VERSION}
```
BryceBeagle: We might want to make this download style consistent with the other build layers. 

I think when I wrote the rest of the Dockerfile, I downloaded directly from PyPI's CDN to avoid the python-pip install, but in hindsight that may have been overkill. I'm not opposed to making every layer just use pip.

Dockerfile
```diff
@@ -95,6 +100,25 @@ RUN sip-install \
...
+################################################################################
+
+FROM build-dep AS sip
+
+# Reuse arguments from previous build scope
+ARG SIP_VERSION
+ARG SIP_ABI_VERSION
+
+# Install SIP including stubs
+# TODO: Find way to build only stubs. This takes way too long
```
BryceBeagle: This TODO comment is probably not relevant as we're not building anything here.

Dockerfile
altendky: Ok, so now that I look the regular PyQt5 package provides its own `.pyi` files.  Why are we building ourselves?  (regardless of how we get the sdist)

Dockerfile
altendky:
```suggestion
```
Agreed.

Dockerfile
BryceBeagle: They didn't used to be: https://github.com/stlehmann/PyQt5-stubs/issues/40
We may be able to avoid building at this point

altendky: I vote we don't mix switching away from building into this right now.  Maybe?  Though I do like consistency.  Turns out the `.pyi` are right in the sdist so I'll switch over to the download/extract/copy structure.  We can revisit not-building separately.

altendky:
@BryceBeagle, thanks for all your feedback and help with my work here.

Dockerfile
altendky: Is the update what you were looking for?

BryceBeagle: I'm not familiar enough with the differences between `find -name` and `find -wholename`, but assuming you had a reason to go with the latter, lgtm

BryceBeagle (approved):

Dockerfile
altendky: Yeah, I knew `-name` was wrong because I had the parent directory in the path but I ran it anyways and it told me to `-wholename`.  At least the present sdist provides two ABIs so we need to pick one over the other.

PR #74, opened 2020-09-23
Provide a separate sip-stubs package

altendky:
Draft for:
- [ ] Self review and testing


altendky:
Err, perhaps I've already reviewed this and realized my silliness...  as it is in fact supposed to be inside PyQt5 isn't it.  \*sigh\* Time for bed.  I'll presumably retract this fully tomorrow.

altendky:
https://github.com/stlehmann/PyQt5-stubs/blob/a878cb8a4b5a2377d519434da7cb9df180cbc4b4/PyQt5-stubs/QtX11Extras.pyi#L24

https://travis-ci.org/github/stlehmann/PyQt5-stubs/builds/729488886#L229
```
PyQt5-stubs/QtX11Extras.pyi:24: error: Cannot find implementation or library stub for module named 'sip'
```

And I went the wrong way here.  `sip` used to be a top level package but now comes from within `PyQt5`.  I'll still leave this until tomorrow but I think the proper change is really to change the import to `from PyQt5 import sip`.

altendky:
Incorrect, see #75 instead.

PR #76, opened 2020-09-23
Ignore complaints about mixed @staticmethod and regular method overloads

altendky:
```
PyQt5-stubs/QtTest.pyi:153: error: Overload does not consistently use the "@staticmethod" decorator on all function signatures.
PyQt5-stubs/QtCore.pyi:4166: error: Overload does not consistently use the "@staticmethod" decorator on all function signatures.
PyQt5-stubs/QtQml.pyi:572: error: Overload does not consistently use the "@staticmethod" decorator on all function signatures.
PyQt5-stubs/QtQml.pyi:583: error: Overload does not consistently use the "@staticmethod" decorator on all function signatures.
PyQt5-stubs/QtNetworkAuth.pyi:242: error: Overload does not consistently use the "@staticmethod" decorator on all function signatures.
PyQt5-stubs/QtMultimedia.pyi:2277: error: Overload does not consistently use the "@staticmethod" decorator on all function signatures.
```


PyQt5-stubs/QtTest.pyi
```diff
@@ -151,9 +151,11 @@ class QTest(sip.simplewrapper):
     @staticmethod
     def mouseClick(window: QtGui.QWindow, button: QtCore.Qt.MouseButton, modifier: typing.Union[QtCore.Qt.KeyboardModifiers, QtCore.Qt.KeyboardModifier] = ..., pos: QtCore.QPoint = ..., delay: int = ...) -> None: ...
     @typing.overload
-    def keySequence(self, widget: QtWidgets.QWidget, keySequence: typing.Union[QtGui.QKeySequence, QtGui.QKeySequence.StandardKey, str, int]) -> None: ...
```
BryceBeagle: I think the two already staticmethods should be removed. There's no static version according to the Qt docs:
https://doc.qt.io/qt-5/qtest.html#keySequence

BryceBeagle (changes requested):
None of these functions now marked as static are actually static. Two of the methods changed to static do seem to have erroneous already-static methods that don't really exist (I left a comment).

See here for current understanding: https://github.com/stlehmann/PyQt5-stubs/blob/master/issues.md#1-mixed-overloads-with-static-and-non-static-methods

altendky:
Well, I definitely glossed over this one.  In my head it was just an odd way to write the method without `self`.  Thanks for catching this and explaining and I'll review more carefully.

BryceBeagle (approved):

BryceBeagle:
We might want to change the `issues.md` to say to use `# type: ignore[misc]` and backport the other changes of this type to add the `misc` too? For consistency.

Issue #60, opened 2020-09-13
What are the classes for modules for?


altendky:
There are a few of these.  What are they for?  I am trying to work through the thousands of errors on master.

https://github.com/stlehmann/PyQt5-stubs/blob/50f45f509d3a03fddb2704b550561f162058ed0b/PyQt5-stubs/Qt3DCore.pyi#L43

altendky:
So I ought have pointed out the issue more directly.  The file is `Qt3DCore` and there is a class inside that is `Qt3DCore` which has all the attributes inside it.  So instead of `Qt3DCore.QAspectEngine` we have `Qt3DCore.Qt3DCore.QAspectEngine`.  So, what do you think?  Just delete that layer?

PR #79, opened 2020-09-23
Remove class layer

altendky:
Ref #60.

Draft for:
- [x] I think I missed all the `Qt3D*` stuff


altendky:
This is basically impossible to review here.  What I at least meant to do was to delete the `class` definition and dedent everything in side.  PyCharm's diff is at least able to visualize this diff well.  I'm not sure what else does...  but not GitHub.

BryceBeagle (approved):
I used this PR as a reason to give the GitHub integration in PyCharm a shot. Pretty nifty! I might start using this more in the future. The diffing worked great on _some_ of the files once I told it to ignore whitespace changes. It did _not_ work great on the files once a line started wrapping at a different point....

I skimmed all the files that didn't diff properly and didn't notice anything egregious. I'll assume this works properly.

PR #81, opened 2020-09-25
Install PyQt5==5.15.* in tox.

altendky:


BryceBeagle:
Is there a way this can just pull the version that matches the version of the stubs being tests? So we don't have to do maintenance here?

altendky:
I considered doing 'something better' but decided to just get it matching for now.  Depending, I may take a pass through CI in general and think about what might work better.  Or we can try to do that sooner than later...

BryceBeagle (approved):

PR #87, opened 2020-09-25
Ignore an inheritance incompatible assignment

altendky:


PyQt5-stubs/QtWidgets.pyi
```diff
@@ -7855,7 +7855,7 @@ class QStyleOptionTab(QStyleOption):
 class QStyleOptionTabV4(QStyleOptionTab):
 
     class StyleOptionVersion(int): ...
```
BryceBeagle, line 7857: I'm really surprised that this was implemented. This seems like a Qt dev-only thing.

This was the only thing I could find mentioning `QStyleOptionTabV4`
https://code.qt.io/cgit/qt/qtbase.git/commit/?id=729f1ab9a0b9fe6ffdefc02df490ec546db38d72

BryceBeagle (approved):

PR #88, opened 2020-09-25
Several # type: ignore[override]

altendky:


PyQt5-stubs/QtPositioning.pyi
```diff
@@ -379,10 +379,8 @@ class QGeoPositionInfoSource(QtCore.QObject):
     def requestUpdate(self, timeout: int = ...) -> None: ...
     def stopUpdates(self) -> None: ...
     def startUpdates(self) -> None: ...
-    @typing.overload
+    # If the error signal is real, it isn't accessible in PyQt 5.15.1
```
BryceBeagle, line 382: Is this a PyQt problem? Should we reach out on mailing list?

BryceBeagle (approved):

PyQt5-stubs/QtPositioning.pyi
altendky: Maybe?  Though I'm not clear how the attribute would be both a method and a signal in C++ either.  I expected to review the diff we end up with against upstream once we feel a bit settled.  But if you want to get started on more upstream fixes sooner, by all means.  `:]`

BryceBeagle: This comment on StackOverflow indicates to me that the signal _is_ accessible at least from C++:
https://stackoverflow.com/a/41006378/8134178

BryceBeagle: I wonder if both are accessible due to the way the C++ Qt `SIGNAL`/`SLOT` "macro" things work

BryceBeagle: I'll probably wait to reach out to upstream until we've got a good changelist for Phil

altendky: So there's a bit more to it, kind of.

```pycon
>>> from PyQt5 import QtPositioning
>>> QtPositioning.QGeoPositionInfoSource.error
<unbound PYQT_SIGNAL error(QGeoPositionInfoSource::Error)>
>>> QtPositioning.QNmeaPositionInfoSource(1).error
<built-in method error of QNmeaPositionInfoSource object at 0x7f798a85a3a0>
>>> QtPositioning.QNmeaPositionInfoSource(1).error()
2
>>> QtPositioning.QNmeaPositionInfoSource(1).error.emit
AttributeError: 'builtin_function_or_method' object has no attribute 'emit'
```

PR #97, opened 2020-10-02
Update changelog after missing many entries in merged PRs

altendky:


BryceBeagle (changes requested):
Not sure how detailed we want to get here, but one I think important enough to include would be something about the _contents_ of `sip.array`

- Added `Sequence`  methods (and `__setitem__`) to `sip.array`

I can see this changelog as a good summary to provide to Phil when we want to work on getting the issues fixed upstream, so maybe it would be best to be pretty detailed?

altendky:
Thanks.  I apparently missed that PR when working through the list.

BryceBeagle (approved):

The-Compiler (commented):
Not going to block this, but why don't format code as code? That'd allow to write ``` `.__getitem__()` ``` rather than `.\_\_getitem\_\_()` and also make things easier to read (IMHO).

altendky:
Because existing entries didn't.  I'll go ahead and submit another PR to update everything since the three of us agree.

PR #99, opened 2020-10-03
Enable strict for mypy

altendky:
Draft for:
- [x] Changelog entry
- [x] Address remaining issues
- [x] Merge fixes from granular PRs for clearer changes and changelog entries
  - https://github.com/stlehmann/PyQt5-stubs/pull/100
  - https://github.com/stlehmann/PyQt5-stubs/pull/101
  - https://github.com/stlehmann/PyQt5-stubs/pull/102
  - https://github.com/stlehmann/PyQt5-stubs/pull/103
  - https://github.com/stlehmann/PyQt5-stubs/pull/104

PR #94, opened 2020-09-29
Annotate OpenMode flag classes with bitwise methods

ghost:
Hi there, this is my contribution to this repo. I was chatting with @altendky via Matrix in the `python/typing` channel and I decided I wanted to help out for fun. This issue looked simple enough to do and didn't look like it was being worked on. I haven't tested these changes since I'm new to using Python and mypy.

If all is good then this closes #52

PR #101, opened 2020-10-04
Remove unneeded ignores on QIODevice.OpenMode

altendky:


altendky:
I'm not clear why this is different now than in #94...  unless this build goes and fails without the other changes in #99.

BryceBeagle (approved):

Issue #11, opened 2019-09-11
Consider using pytest-mypy-plugins


sobolevn:
Hi! Thanks for this awesome project!

I am [TypedDjango](https://github.com/TypedDjango) team member, we maintain types for, well, django. And we do pretty much the same job.

For example, we also test our types the similar way as you do in [`tests/`](https://github.com/stlehmann/PyQt5-stubs/tree/master/tests). We even created a tool called [`pytest-mypy-plugins`](https://github.com/typeddjango/pytest-mypy-plugins) ([announcing post](https://sobolevn.me/2019/08/testing-mypy-types)) to help us with this task. Maybe it will be also helpful to you as well. 

That's how the simplest test looks like:

```yaml
- case: compose_two_wrong_functions
  main: |
    from returns.functions import compose

    def first(num: int) -> float:
        return float(num)

    def second(num: float) -> str:
        return str(num)

    reveal_type(compose(first, second)(1))  # N: builtins.str*
```

Ask any questions you have! 

P.S. This project was listed in [awesome-python-stubs](https://github.com/typeddjango/awesome-python-stubs) list.

stlehmann:
@sobolevn Thanks for the hint. I will have a closer look at pytest-mypy-plugins for sure.

bluebird75:
I had a look at the plugin. I find the syntax a bit complex and I prefer the simple Python code which we have right now for validating our stubs in tests.

If nobody is planning to use this plugin in PyQt5-stubs, we can close this issue.

PR #102, opened 2020-10-04
fix `pyqtSlot` `result` parameter type and `Callable` generic

altendky:


PyQt5-stubs/QtCore.pyi
```diff
@@ -9232,7 +9232,7 @@ QT_VERSION = ... # type: int
 QT_VERSION_STR = ... # type: str
 
 
-FuncT = typing.TypeVar('FuncT', bound=typing.Callable)  # For a correct pyqtSlot annotation
+FuncT = typing.TypeVar('FuncT', bound=typing.Callable[..., typing.Any])  # For a correct pyqtSlot annotation
```
BryceBeagle: What is this change for?

https://docs.python.org/3/library/typing.html#typing.Callable
> A plain Callable is equivalent to Callable[..., Any]

tests/pyqtslot.py
```diff
@@ -1,7 +1,12 @@
 from PyQt5.QtCore import pyqtSlot
 
 @pyqtSlot(str)
-def func(s: str) -> int:
+def func_none(s: str) -> None:
```
BryceBeagle, line 4: Do we want to add some tests that force/catch an exception?

BryceBeagle (changes requested):

tests/pyqtslot.py
altendky: What exceptions do you have in mind?  Or do you mean type hint failures?

PyQt5-stubs/QtCore.pyi
altendky: When strict, mypy doesn't like unparametrized generics.

PyQt5-stubs/QtCore.pyi
```diff
@@ -9264,7 +9264,7 @@ def oct_(s: QTextStream) -> QTextStream: ...
 def bin_(s: QTextStream) -> QTextStream: ...
 def Q_RETURN_ARG(type: typing.Any) -> QGenericReturnArgument: ...
 def Q_ARG(type: typing.Any, data: typing.Any) -> QGenericArgument: ...
-def pyqtSlot(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Optional[str] = ...) -> typing.Callable[[FuncT], FuncT]: ...
+def pyqtSlot(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Type[object] = ...) -> typing.Callable[[FuncT], FuncT]: ...
```
altendky: How do you feel about the `typing.Type[object]`.  I starting with `Type` because I was trying to relate it to the function return type but...  I didn't get that working.  Maybe this should just be `object`/`Any`?

BryceBeagle: https://www.riverbankcomputing.com/static/Docs/PyQt5/signals_slots.html

Just noticed we're missing the `revision` argument

BryceBeagle: Could `FuncT` be made generic? I _think_ this would work
```python
T = TypeVar("T")
FuncT = typing.TypeVar('FuncT', bound=typing.Callable[..., T])
def pyqtSlot(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Type[T] = ...) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
```

However, looking at the docs:
> result – the type of the result and may be a Python type object or a string that specifies a C++ type

Not sure how the C++ string would interact with this. Maybe
```python
def pyqtSlot(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Union[typing.Type[T], str] = ...) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
```

BryceBeagle: Just noticed that `types` has the same comment in the docs about "may be a Python type object or a string", so maybe we can be more specific than `typing.Any`

BryceBeagle (changes requested):

tests/pyqtslot.py
BryceBeagle: Yeah. Tests cases that raise mypy errors to demonstrate that the annotations are correct.

altendky: I didn't expect to get into that until a better testing setup is worked on.  Perhaps what is mentioned in #11.  I haven't dug much into that yet.

altendky:
Do you think we should try to do more here now?

BryceBeagle:
Did you see my comment about making `FuncT` generic?

PyQt5-stubs/QtCore.pyi
altendky: Did you mean "can't be more specific"?  When I read this I thought you made a comment then backed away from it.  I did make an effort to relate the `result` parameter to the return type of the `typing.Callable` parameter and return but failed.  I could make another attempt for you to look at.

BryceBeagle: No those were two different comments. I was suggesting that we change `result` to use a generic. Then I noticed that `types` is documented in a similar fashion and remarked that we could maybe do better than its current `typing.Any`

altendky: For the `types`, they can be any types including user defined and they don't necessarily even inherit from type what with metatypes, right?  (plus `str`)  So the only thing I can think of is to try to relate it with the parameters of the wrapped callable, but mypy doesn't do that yet.  There's some related PEP but I can't find it at the moment.  Do you have something specific in mind here?

I'll try what you shared though it looks familiar.

BryceBeagle: Yeah you're probably right that `types` is as good as mypy currently allows

altendky: Here's yours along with the existing and a `Protocol` attempt.

https://mypy-play.net/?mypy=latest&python=3.8&gist=314620351fccbbfd297833d345b40a6d
```python
import typing


FuncT = typing.TypeVar('FuncT', bound=typing.Callable[..., typing.Any])

def pyqtSlot(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Any = ..., revision: int = ...) -> typing.Callable[[FuncT], FuncT]: ...


T = typing.TypeVar("T")
FuncT2 = typing.TypeVar('FuncT2', bound=typing.Callable[..., T])
def pyqtSlot2(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Type[T] = ...) -> typing.Callable[[FuncT2[T]], FuncT2[T]]: ...


T_covariant = typing.TypeVar("T_covariant", covariant=True)
class P(typing.Protocol, typing.Generic[T_covariant]):
    def __call__(self, *args: typing.Any, **kwargs: typing.Any) -> T_covariant: ...


def pyqtSlot3(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Type[T_covariant] = ...) -> typing.Callable[[P[T_covariant]], P[T_covariant]]: ...


def f(s: str, i: int) -> float: ...


reveal_type(pyqtSlot(result=float)(f))
reveal_type(pyqtSlot2(result=float)(f))
reveal_type(pyqtSlot3(result=float)(f))
```
```
main.py:11: error: Type variable "FuncT2" used with arguments
main.py:25: note: Revealed type is 'def (s: builtins.str, i: builtins.int) -> builtins.float'
main.py:26: error: Value of type variable "FuncT2" of function cannot be "Callable[[str, int], float]"
main.py:26: note: Revealed type is 'def (s: builtins.str, i: builtins.int) -> builtins.float'
main.py:27: note: Revealed type is 'main.P[builtins.float*]'
main.py:27: error: Argument 1 has incompatible type "Callable[[str, int], float]"; expected "P[float]"
Found 3 errors in 1 file (checked 1 source file)
```

I don't think we get to make a generic typevar.

BryceBeagle: Is there a reason we need to make `FuncT` a TypeVar?

```python
import typing

S = typing.TypeVar("S")
T = typing.TypeVar("T")
FuncT = typing.Callable[..., S]

def pyqtSlot(*types: typing.Any, name: typing.Optional[str] = ..., result: typing.Type[T] = ...) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...


def f(s: str, i: int) -> float: ...


reveal_type(pyqtSlot(result=float)(f))
```

```
main.py:13: note: Revealed type is 'def (*Any, **Any) -> builtins.float*'
```

```python
def f(s: str, i: int) -> float: ...


reveal_type(pyqtSlot(result=bool)(f))
```
```
main.py:13: note: Revealed type is 'def (*Any, **Any) -> builtins.bool*'
main.py:13: error: Argument 1 has incompatible type "Callable[[str, int], float]"; expected "Callable[..., bool]"
Found 1 error in 1 file (checked 1 source file)
```

altendky: You win (I think), thanks.

BryceBeagle:
Sorry not trying to drag this PR on and on again, but I think we need this for the overloads:

```python
@typing.overload
def pyqtSlot(*types: typing.Any) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, result: typing.Union[typing.Type[T], str]) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str, result: typing.Union[typing.Type[T], str]) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str, revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, result: typing.Union[typing.Type[T], str], revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str, result: typing.Union[typing.Type[T], str], revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
```

What I changed:
* Added overloads for each combination of all three optional arguments. Note that they're not optional... just overloads *. This is pretty verbose, but I don't know of a better option.
* Removed the Optional-ality of the `str` arguments
* Removed `...` defaults from arguments because they don't actually have defaults
* Combined the `result` types into a Union because the Union-ization doesn't affect the return type/other arguments

\* Note: This contradicts the documentation [here](https://www.riverbankcomputing.com/static/Docs/PyQt5/signals_slots.html#the-pyqtslot-decorator) where the later arguments are in ever-nesting brackets. The documentation implies that you can't use `revision` without using `result` but this doesn't appear to be the case. Some of the examples in the docs even contradict the specification. 

```pycon
>>> pyqtSlot(int, name=None)
```
```
TypeError: pyqtSlot() argument 1 must be str, not None
```
Side note: It looks like `argument 1` here is a bugged miscount on PyQt's side, as this is argument 2.

This is _further_ contradicted `pyqtSlot.__doc__`:
```
@pyqtSlot(*types, name: Optional[str], result: Optional[str])

This is a decorator applied to Python methods of a QObject that marks them
as Qt slots.
The non-keyword arguments are the types of the slot arguments and each may
be a Python type object or a string specifying a C++ type.
name is the name of the slot and defaults to the name of the method.
result is type of the value returned by the slot.
```

We should probably mention this to Phil too. 

This is a mess :facepalm:

altendky:
You can't use `Union` there with `--strict` because it makes the `TypeVar` be `<nothing>`.  I'll expand and commit.

https://mypy-play.net/?mypy=latest&python=3.8&flags=strict&gist=9608229557029541b9a90553bce77414

```python
import typing


T = typing.TypeVar("T")
FuncT = typing.Callable[..., T]
@typing.overload
def pyqtSlot(*types: typing.Any) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, result: typing.Union[typing.Type[T], str]) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str, result: typing.Union[typing.Type[T], str]) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str, revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, result: typing.Union[typing.Type[T], str], revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...
@typing.overload
def pyqtSlot(*types: typing.Any, name: str, result: typing.Union[typing.Type[T], str], revision: int) -> typing.Callable[[FuncT[T]], FuncT[T]]: ...

@pyqtSlot(str, result='int')
def func_str(s: str) -> int:
    return 42
```
```
main.py:6: error: An overloaded function outside a stub file must have an implementation
main.py:23: error: Argument 1 has incompatible type "Callable[[str], int]"; expected "Callable[..., <nothing>]"
```

BryceBeagle (approved):

BryceBeagle (changes requested):
Actually we probably also want to update the changelog too

BryceBeagle:
```
fix `pyqtSlot` parameter typing and overloads
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * [#51](https://github.com/stlehmann/PyQt5-stubs/pull/51) adds `pyqtBoundSignal.signal` hinted as `str`
 
 ### Changed
+* [#102](https://github.com/stlehmann/PyQt5-stubs/pull/102) fix `pyqtSlot` `result` parameter type to `typing.Type[object]`
```
altendky:
```suggestion
* [#102](https://github.com/stlehmann/PyQt5-stubs/pull/102) fix `pyqtSlot` parameter typing and overloads
```

Are you familiar with the suggestion feature?  I can't remember if you had used it.  It gives me a button to create a commit and includes you as the author (or both of us, or something).

BryceBeagle: I know it exists, but I couldn't find it in the 5 seconds I looked for it. I'll try to use it in the future.

BryceBeagle (approved):

PR #103, opened 2020-10-04
`pyqtBoundSignal.disconnect()`'s `slot` parameter is optional

altendky:


PyQt5-stubs/QtCore.pyi
```diff
@@ -35,7 +35,7 @@ class pyqtBoundSignal:
 
     def emit(self, *args: typing.Any) -> None: ...
     def connect(self, slot: "PYQT_SLOT") -> "QMetaObject.Connection": ...
-    def disconnect(self, slot: typing.Union["PYQT_SLOT", "QMetaObject.Connection"]=None) -> None: ...
+    def disconnect(self, slot: typing.Optional[typing.Union["PYQT_SLOT", "QMetaObject.Connection"]]=None) -> None: ...
```
BryceBeagle: This isn't an Optional argument. We actually need to add an `overload` here

```python
class Test(QObject):
    a = pyqtSignal()

    @pyqtSlot()
    def blah(self):
        print("Here")

b = Test()
b.a.connect(b.blah)
b.a.disconnect(b.blah) 
b.a.disconnect(None)
```
```pytb
  File "/home/ignormies/.config/JetBrains/PyCharm2020.2/scratches/scratch.py", line 13, in <module>
    b.a.disconnect(None)
TypeError: argument 1 has unexpected type 'NoneType'
```

BryceBeagle (changes requested):

PyQt5-stubs/QtCore.pyi
altendky: Quite so, thanks.

```suggestion
    @typing.overload
    def disconnect(self) -> None: ...
    @typing.overload
    def disconnect(self, slot: typing.Union["PYQT_SLOT", "QMetaObject.Connection"]) -> None: ...
```

BryceBeagle (approved):

PR #104, opened 2020-10-04
sip.voidptr handles integer values and sequences

altendky:


PyQt5-stubs/sip.pyi
```diff
@@ -63,7 +63,7 @@ class array(Sequence[_T], Generic[_T]):
 # The voidptr type.
 class voidptr:
 
-    def __init__(addr: Union[int, Buffer], size: int = -1, writeable: bool = True) -> None: ...
+    def __init__(addr: Union[int, None, bytes, bytearray, 'voidptr'], size: int = -1, writeable: bool = True) -> None: ...
```
BryceBeagle: Is this `__init__` somehow static?

BryceBeagle (changes requested):

PyQt5-stubs/sip.pyi
altendky: Heh, that's fun.  Let's see how this works out.

```suggestion
    def __init__(self, addr: Union[int, None, bytes, bytearray, 'voidptr'], size: int = -1, writeable: bool = True) -> None: ...
```

BryceBeagle (changes requested):
I'm not entirely convinced that `voidptr.__init__` doesn't have some magic going on behind the scenes allowing it to have a `self`-less ` __init__`. Maybe it does something very strange with CPython, or with `__new__`.

Also, I tried looking through the `sip` source for this class and I have no idea where it is, or how it's generated, so that was a bust.

I think we should document this change so we can ask about it in the future. Can it be added to the changelog?

altendky:
Without claiming that I now think I know what's up...

```pycon
>>> help(sip.voidptr)
Help on class voidptr in module PyQt5.sip:

class voidptr(builtins.object)
 |  Methods defined here:
 |  
 |  __bool__(self, /)
 |      self != 0
 |  
 |  __delitem__(self, key, /)
 |      Delete self[key].
 |  
 |  __getitem__(self, key, /)
 |      Return self[key].
 |  
 |  __int__(self, /)
 |      int(self)
 |  
 |  __len__(self, /)
 |      Return len(self).
 |  
 |  __setitem__(self, key, value, /)
 |      Set self[key] to value.
 |  
 |  asarray(...)
 |  
 |  ascapsule(...)
 |  
 |  asstring(...)
 |  
 |  getsize(...)
 |  
 |  getwriteable(...)
 |  
 |  setsize(...)
 |  
 |  setwriteable(...)
 |  
 |  ----------------------------------------------------------------------
 |  Static methods defined here:
 |  
 |  __new__(*args, **kwargs) from builtins.type
 |      Create and return a new object.  See help(type) for accurate signature.
```

```pycon
>>> inspect.signature(sip.voidptr.__init__)
<Signature (self, /, *args, **kwargs)>
```

```pycon
>>> sip.voidptr.__init__()
TypeError: descriptor '__init__' of 'object' object needs an argument
```

If we end up merging something here without full satisfaction, maybe it would make sense to have a ticket rather than a changelog entry to track the need for future work?

BryceBeagle:
> If we end up merging something here without full satisfaction, maybe it would make sense to have a ticket rather than a changelog entry to track the need for future work?

If we continue with the change as-is, we should probably include it in the changelog. Opening an issue is a good idea regardless of if we continue with the change here in this PR

altendky:
* [#104](https://github.com/stlehmann/PyQt5-stubs/pull/104) `sip.voidptr` handles integer values and sequences

Does it need `` ' and takes `self`' `` added on the end?

I thought you suggested more changelog commentary covering uncertainty about what this bit of stub should be.  That bit I thought belonged in an issue not a changelog entry.  I agree the change made here ought to have an entry for the change itself.

BryceBeagle:
> Does it need ' and takes `self`' added on the end?

Yeah that's what I'm suggesting

BryceBeagle (approved):

PR #107, opened 2020-10-08
(WIP) Update the parameter type annotations for pyqtSignal.__init__

### LINKED ISSUE #105: pyqtSignal needs the same treatment as pyqtSlot
`pyqtSignal.__init__` has bunch of parameters and overloads that are not properly annotated

https://www.riverbankcomputing.com/static/Docs/PyQt5/signals_slots.html#PyQt5.QtCore.pyqtSignal

https://github.com/stlehmann/PyQt5-stubs/blob/811462b34ee151b898289ae8f1de8af30c690c55/PyQt5-stubs/QtCore.pyi#L44-L45

BryceBeagle:
Resolves #105 

The *types annotation is currently broken. Feedback appreciated

```pytb
test_stubs.py::test_stubs[pyqtsignal.py]
pyqtsignal.py:6: error: Value of type variable "_SignalTypesT" of "pyqtSignal" cannot be "List[Type[str]]"
pyqtsignal.py:8: error: Value of type variable "_SignalTypesT" of "pyqtSignal" cannot be "object"
```

Minimal reprduction of issue: 
https://mypy-play.net/?mypy=latest&python=3.8&gist=57db3e86d7202dc8ebe4f24965409a3f

Edit: Opened an issue on the mypy issue tracker: https://github.com/python/mypy/issues/9569


PyQt5-stubs/QtCore.pyi
```diff
@@ -40,9 +40,22 @@ class pyqtBoundSignal:
     @typing.overload
     def disconnect(self, slot: typing.Union["PYQT_SLOT", "QMetaObject.Connection"]) -> None: ...
 
-
+_SignalTypesT = typing.TypeVar("_SignalTypesT", type, typing.List[type])
```
altendky, line 43:
```suggestion
_SignalTypesT = typing.TypeVar("_SignalTypesT", typing.Union[type, typing.List[type]])
```
I can't say I know the reasons for why this works and the separate types don't but...  it does.  `:|`  Is there something that makes it wrong despite passing?

altendky: Err, hold that thought...  maybe I didn't test properly here...

altendky: Ok, trying again and hoping to be less silly.  As is, there is no relationship being created between multiple hints so this can just be an alias instead, I think?

```suggestion
_SignalTypesT = typing.Union[type, typing.List[type]]
```
(except with a different variable name changed here and at the use sites)

BryceBeagle: The issue here is that I need to keep all of the var args of the same type. Using the Union allows for
```python
a = pyqtSignal([int, bool], float)
```
but this is not allowed by PyQt. You can use _only_ types OR _only_ lists of types

altendky: Well, I think my second attempt was _less_ silly...  though still not helpful.  Humm.

PR #110, opened 2020-11-08
Fixed QItemSelection.selectionChanged signature

TilmanK:


altendky (approved):
Matches the other uses, thanks.

PR #112, opened 2020-12-03
Correct QApplication.setStyle() to allow str as argument type.

TilmanK:
Updates QApplication.setStyle() to also allow str as arguments (https://doc.qt.io/qt-5/qapplication.html#setStyle-1).


PyQt5-stubs/QtWidgets.pyi
```diff
@@ -389,7 +389,8 @@ class QWidget(QtCore.QObject, QtGui.QPaintDevice):
     def setDisabled(self, a0: bool) -> None: ...
     def setEnabled(self, a0: bool) -> None: ...
     def isEnabledTo(self, a0: 'QWidget') -> bool: ...
-    def setStyle(self, a0: 'QStyle') -> None: ...
+    def setStyle(self, style: 'QStyle') -> None: ...
+    def setStyle(self, style: str) -> None: ...
```
altendky: I think this would fix the error in the build.

```suggestion
    @typing.overload
    def setStyle(self, style: 'QStyle') -> None: ...
    @typing.overload
    def setStyle(self, style: str) -> None: ... 
```

But...  I think this was applied to `QWidget` not `QApplication`?

altendky (changes requested):

PyQt5-stubs/QtWidgets.pyi
BryceBeagle: Would a `Union` not be more applicable here?

altendky: Maybe?  Seems reasonable, I think.

TilmanK: You guys are right. The Union is the way to go. Didn't notice the double setStyle definition at first...

BryceBeagle: As @altendky mentioned, you're still modifying `QWidget.setStyle`, which [doesn't have any overloads](https://doc.qt.io/qt-5/qwidget.html#setStyle). You'll need to change `QApplication.setStyle`, which is what you linked and has the overloads.

BryceBeagle: Actually, it already has that overload: 

https://github.com/stlehmann/PyQt5-stubs/blob/704207e90bee7b36ec9861dfa6b39f06a27c6718/PyQt5-stubs/QtWidgets.pyi#L1189-L1193

And I was wrong about using a `Union` because the return type differs between the two signatures.

altendky:
@TilmanK, did we conclude that this submission was based on some confusion?  Or is there an issue still to be fixed?

TilmanK:
@altendky Sorry for the delayed response Kyle. Yes, you are right. I just checked the code and indeed, the implementation was correct, and I was fiddling around in the wrong place. Sorry.

PR #113, opened 2020-12-04
Merge upstream 5.15.2

altendky:


Dockerfile
```diff
@@ -1,16 +1,16 @@
 # Careful, the version and build date are both dates, but different formats
-ARG ARCH_VERSION="20200908"
-ARG BUILD_DATE="2020/09/18"
+ARG ARCH_VERSION="base-20201129.0.10056"
```
BryceBeagle, line 5: I supposed the comment above is no longer necessary now that the tag changed formats. It should probably just explain that this is used as a tag from [dockerhub](https://hub.docker.com/_/archlinux?tab=tags&page=1&ordering=last_updated)

BryceBeagle (approved):

Dockerfile
altendky: Updated, thanks for pointing that out.

PR #118, opened 2020-12-19
Setup for GitHub Actions

altendky:


altendky:
Travis is slow and this gets us other platforms with good Python support.

https://www.traviscistatus.com/#month
image

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+    branches:
+      - main
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - "*"
+  schedule:
+    # Daily at 06:14
+    - cron: '14 6 * * *'
```
BryceBeagle: Any reason for this exact time? Do we really need to it run daily anyways?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+  all:
+    name: All
+    runs-on: ubuntu-latest
+    needs:
+      - test
+    steps:
+      - name: This
+        shell: python
+        run: |
+          import this
```
BryceBeagle, line 80: What's the purpose of this?

BryceBeagle (changes requested):

.github/workflows/ci.yml
altendky: Each time I copy this I shift the start time a bit just to stagger.  I probably don't stagger sufficiently but...  The general intent is sometime in the night.  This project probably benefits less from nightly builds but it is nice to know when things broke and building more often makes it more likely you will know which one thing is the cause of a break.  But, meh, not a big deal.

.github/workflows/ci.yml
altendky: When you tell GitHub what checks are required you have to check individual jobs, not the workflow.  This All job gives you a single job to select there.  For example, if we add another python and don't do this then we have to go check three more jobs and do so in a way that will make other PRs fail until that change is merged into them etc.  With this it is (more) contained in the repo.

The `import this` is just a fast thing to put in the step that might occasionally make someone smile.

.github/workflows/ci.yml
BryceBeagle: It might be best to at least comment that the exact time has no significance, so that it doesn't confuse someone in the future

.github/workflows/ci.yml
BryceBeagle: Not sure I understand why an `All` job is important? If we add another Python, wouldn't we just add it to the build matrix?

altendky: If that's all you do then it won't be marked as required and you will be allowed to merge a PR even when it fails.  The images below are from QTrio but show how I don't have to check every single job created by the matrix and keep those checks in sync with (one of the branches of) the code.  Note how the `All` job `needs:` the `test` job and thus is only successful (only even runs) if every matrixed 'sub-job' of `test` succeeds.

The alternative is that when you change the matrix (either addition or subtraction of sub-jobs (cells?  I don't know...)) you also go into the GitHub project settings and change the required jobs.  One issue is that you have to do this at all.  The other issue is that when you do that then all other open PR's are failing.  It is a global setting despite the CI config and the actually defined jobs being a branch setting.

This is certainly not required, most projects do not do this.  But, mine do.  `:]`  In this case we don't control the required jobs anyways until we get PyQt5-stubs moved into an org, so it kind of doesn't matter until then.  We're just going to have to look for ourselves before merging regardless.

image
image
image

BryceBeagle: Sounds good to me. Thanks for the explanation

BryceBeagle (approved):

PR #121, opened 2020-12-23
Stop testing macOS and Windows for now

altendky:
We can introduce these after stubtests and dealing with each OS separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,7 @@ jobs:
           - name: Linux
             runs-on: ubuntu-latest
             matrix: linux
+#          https://github.com/stlehmann/PyQt5-stubs/issues/122
 #          - name: Windows
 #            runs-on: windows-latest
 #            matrix: windows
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -25,12 +25,12 @@ jobs:
           - name: Linux
             runs-on: ubuntu-latest
             matrix: linux
-          - name: Windows
-            runs-on: windows-latest
-            matrix: windows
-          - name: macOS
-            runs-on: macos-latest
-            matrix: macos
+#          - name: Windows
```
BryceBeagle, line 29:
```suggestion
#          https://github.com/stlehmann/PyQt5-stubs/issues/122
#          - name: Windows
```

BryceBeagle (changes requested):

altendky:
Yeah, I was being lazy.  `:[`  Was going to create re-enabling PRs right after merging this then the stubtest.  Thanks.

BryceBeagle (approved):

Issue #122, opened 2020-12-23
Add Windows and Mac tests back


BryceBeagle:
Stub testing had issues with these platforms: #95 
Disabled temporarily in #121

bluebird75:
We can close it now

PR #123, opened 2020-12-23
Enable testing on Windows, and make them pass

altendky:
Partially addresses #122

Draft for:
- [ ] Tests actually passing


bluebird75:
This one can be closed, we now have tests on Windows running fine. The branch windows can also probably be deleted

PR #124, opened 2020-12-23
Enable testing on macOS, and make them pass

altendky:
Partially addresses #122

Draft for:
- [x] Tests actually passing


bluebird75 (approved):
awesome

PR #125, opened 2020-12-31
QtWidgets.QMessageBox enumerations are available in two places...

altendky:


altendky:
@BryceBeagle, this seems really bad...  but since they are both places?  Or maybe we would rather just encourage people to use the new location?  (they are all attributes of the enumeration itself in pyqt6, right?)

```pycon
Python 3.9.0 (default, Dec 18 2020, 14:15:28) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from PyQt5 import QtWidgets
>>> QtWidgets.QMessageBox.NoIcon
0
>>> QtWidgets.QMessageBox.Icon.NoIcon
0
>>> QtWidgets.QMessageBox.InvalidRole
-1
>>> QtWidgets.QMessageBox.ButtonRole.InvalidRole
-1
>>> QtWidgets.QMessageBox.NoButton
0
>>> QtWidgets.QMessageBox.StandardButton.NoButton
0
```

Observed while getting:
```
qtrio/examples/readme/qt.py:17: error: "Type[Icon]" has no attribute "Question"  [attr-defined]
```
From:
https://github.com/altendky/qtrio/blob/1d969f8ec5be7bd5dc7b9c3f38e344f5a46e41c5/qtrio/examples/readme/qt.py#L16-L21
```python
    return QtWidgets.QMessageBox(
        QtWidgets.QMessageBox.Icon.Question,
        "Hello",
        "",
        QtWidgets.QMessageBox.Ok,
    )
```

BryceBeagle (approved):
I think the stubs should accurately reflect the library, even if the library does something strange.

Are other Enums in the library available like this? I recall a conversation on the mailing list where PyQt6 was switching to only scoped Enums. 

The docs for PyQt 5.11 mention this:
https://www.riverbankcomputing.com/static/Docs/PyQt5/gotchas.html#enums

altendky:
Sounds like this will apply to all old enums?  And new enums will be available only as attributes of the enum?  We'll see how far Phil makes it following his query on the list about how to check the official hints.  We also need to figure out how we're going to approach PyQt6...  Though I was kind of hoping to put off dealing with Qt6 altogether for a few months.  `:|`

PR #129, opened 2021-01-06
Correct annotations for signals in QThread and QNetworkAccessManager

TilmanK:


PyQt5-stubs/QtNetwork.pyi
```diff
@@ -764,13 +764,13 @@ class QNetworkAccessManager(QtCore.QObject):
     def setProxyFactory(self, factory: 'QNetworkProxyFactory') -> None: ...
     def proxyFactory(self) -> 'QNetworkProxyFactory': ...
     def createRequest(self, op: 'QNetworkAccessManager.Operation', request: 'QNetworkRequest', device: typing.Optional[QtCore.QIODevice] = ...) -> 'QNetworkReply': ...
-    def preSharedKeyAuthenticationRequired(self, reply: 'QNetworkReply', authenticator: 'QSslPreSharedKeyAuthenticator') -> None: ...
+    preSharedKeyAuthenticationRequired = typing.ClassVar[QtCore.pyqtSignal]
     def networkAccessibleChanged(self, accessible: 'QNetworkAccessManager.NetworkAccessibility') -> None: ...
```
BryceBeagle: [`networkAccessibleChanged`](https://doc.qt.io/qt-5/qnetworkaccessmanager-obsolete.html#networkAccessibleChanged) is a signal too, just deprecated/obsolete. I think we should fix it as well

BryceBeagle (changes requested):
Can you add an entry to https://github.com/stlehmann/PyQt5-stubs/blob/b8e0a2b482de03a97b74af33006ffa13e13965de/CHANGELOG.md as well?

PyQt5-stubs/QtNetwork.pyi
TilmanK: You're right, haven't thought about deprecated signals.

BryceBeagle (approved):
I think this PR looks good, however you may want to check your git config to make sure you're committing with the credentials that GitHub expects

TilmanK:
Hey, is there anything else you need for this PR? You'll squash it anyway, so the credentials won't matter, right?

altendky:
@TilmanK, I personally prefer to do a regular merge generally.  If you would prefer a squash to get rid of the commit name/email, we can.

TilmanK:
Ah, sorry! I totally forgot that I wanted to rebase the commits. Anyway, next time then...
Thanks for accepting the PR.

PR #130, opened 2021-02-16
The primary branch here is master

altendky:


BryceBeagle (approved):
Alternatively, should we just follow the new GitHub ~standard and [rename the primary branch](https://github.com/github/renaming#renaming-existing-branches) to `main`? I have no strong opinions either way, though

altendky:
New projects I make use `main` and I'm fine switching over.  I hadn't looked back since the GitHub recommendation was to wait until they finished their tooling.  I probably won't prioritize the transition soon personally but if you want to that'd be great.  Maybe post release though.

PR #132, opened 2021-02-18
Update project URL in setup.py

altendky:


BryceBeagle (approved):
Do we change `author` and `author_email` too?

altendky:
Author wouldn't change unless requested by the present author.  We could add maintainers.  Let's shift this to #133 to elevate it.

PR #137, opened 2021-02-25
Build 5.15.3

altendky:
Draft for:
- [x] Self review
- [x] Using the current SIP
- [x] Explaining the weird extra manually installation to help the Docker not fail cryptically https://github.com/python-qt-tools/PyQt5-stubs/pull/137/files#r583023429
- [x] ~~Maybe use a not-CN Arch CDN~~


Dockerfile
```diff
@@ -26,6 +26,10 @@ FROM archlinux:${ARCH_VERSION} AS build-dep
 # Reuse argument from previous build scope
 ARG BUILD_DATE
 
+RUN patched_glibc=glibc-linux4-2.33-4-x86_64.pkg.tar.zst && \
+    curl -LO "https://repo.archlinuxcn.org/x86_64/$patched_glibc" && \
+    bsdtar -C / -xvf "$patched_glibc"
```
BryceBeagle, line 35: Why Archlinux CN? Is this related to https://bugs.archlinux.org/task/69563?

altendky: Yeah, that looks like the one, sorry I didn't add the comment straight away.

BryceBeagle: But why the China CDN?

altendky: Because that's what got copy/pasted from Qutebrowser. :]  I'll add changing that to the list.

https://github.com/qutebrowser/qutebrowser/blob/30d54c8da4a8e091dbe439770d4e1796dc7c78dc/scripts/dev/ci/docker/Dockerfile.j2#L3-L8
```dockerfile
# WORKAROUND for glibc 2.33 and old Docker
# See https://github.com/actions/virtual-environments/issues/2658
# Thanks to https://github.com/lxqt/lxqt-panel/pull/1562
RUN patched_glibc=glibc-linux4-2.33-4-x86_64.pkg.tar.zst && \
    curl -LO "https://repo.archlinuxcn.org/x86_64/$patched_glibc" && \
    bsdtar -C / -xvf "$patched_glibc"
```

altendky: I don't know my way around arch but it appears that `archlinuxcn.org` isn't just a mirror but rather may be the origin of this file.  Are you familiar enough to know otherwise?

https://wiki.archlinux.org/index.php/Unofficial_user_repositories_(%E7%AE%80%E4%BD%93%E4%B8%AD%E6%96%87)#archlinuxcn

BryceBeagle: You're right. I just assumed that it was a mirror.

BryceBeagle (approved):
These are just the upstream-generated stubs without any of our changes, right?

altendky:
That's the intent, yes.

PR #138, opened 2021-02-26
Merge upstream 5.15.3

altendky:
Draft for:
- [x] Actually merging #137
- [x] Update merge from the `upstream` branch to here
- [x] Self review of the net changes.
- [x] Tests passing
- [x] Changelog entry


altendky:
I'm not exactly sure how I feel about dropping testing for 3.5.  But, we won't be getting 5.15.3 in 3.5 and we won't pass stubtest in both 5.15.3 and 5.15.2 (or probably really any two versions since some interface presumably changed).  So, I'm not really sure what it would mean to support multiple versions.  If it were a Python version issue you can do some amount of `if sys.version_info >= (3, 7):` and such.  But for the PyQt version I think we would have to either make a plugin or use `--always-true PYQT_5_15_2` or such and then our users would have to configure the same when running Mypy.  And that doesn't start into stubtest either...

BryceBeagle:
3.5 [has reached EOL](https://www.python.org/downloads/release/python-3510/). I'm fine dropping support for it

PyQt5-stubs/Qt3DCore.pyi
```diff
@@ -39,17 +39,17 @@ PYQT_OPENGL_BOUND_ARRAY = typing.Union[typing.Sequence[int],
...
+    NodeCreated = ... # type: ChangeFlag
+    NodeDeleted = ... # type: ChangeFlag
+    PropertyUpdated = ... # type: ChangeFlag
+    PropertyValueAdded = ... # type: ChangeFlag
+    PropertyValueRemoved = ... # type: ChangeFlag
+    ComponentAdded = ... # type: ChangeFlag
+    ComponentRemoved = ... # type: ChangeFlag
+    CommandRequested = ... # type: ChangeFlag
+    CallbackTriggered = ... # type: ChangeFlag
+    AllChanges = ... # type: ChangeFlag
```
BryceBeagle, line 52: Missing module-level versions

PyQt5-stubs/QtCore.pyi
```diff
@@ -61,13 +68,13 @@ PYQT_SLOT = typing.Union[typing.Callable[..., object], pyqtBoundSignal]
...
-QtFatalMsg = ... # type: QtMsgType
-QtSystemMsg = ... # type: QtMsgType
-QtInfoMsg = ... # type: QtMsgType
+class QtMsgType(int):
+    QtDebugMsg = ... # type: QtMsgType
+    QtWarningMsg = ... # type: QtMsgType
+    QtCriticalMsg = ... # type: QtMsgType
+    QtFatalMsg = ... # type: QtMsgType
+    QtSystemMsg = ... # type: QtMsgType
+    QtInfoMsg = ... # type: QtMsgType
```
BryceBeagle, line 77: Missing module-level versions

PyQt5-stubs/QtWebKit.pyi
```diff
@@ -59,10 +59,10 @@ class QWebDatabase(sip.simplewrapper):
...
 class QWebElement(sip.simplewrapper):
 
-    class StyleResolveStrategy(int): ...
-    InlineStyle = ... # type: 'QWebElement.StyleResolveStrategy'
-    CascadedStyle = ... # type: 'QWebElement.StyleResolveStrategy'
-    ComputedStyle = ... # type: 'QWebElement.StyleResolveStrategy'
+    class StyleResolveStrategy(int):
+        InlineStyle = ... # type: QWebElement.StyleResolveStrategy
+        CascadedStyle = ... # type: QWebElement.StyleResolveStrategy
+        ComputedStyle = ... # type: QWebElement.StyleResolveStrategy
```
BryceBeagle, line 65: I would assume these are missing from the containing class too.

PyQt5-stubs/QtWebKit.pyi
```diff
@@ -266,9 +266,9 @@ class QWebPluginFactory(QtCore.QObject):
 
 class QWebSecurityOrigin(sip.simplewrapper):
 
-    class SubdomainSetting(int): ...
-    AllowSubdomains = ... # type: 'QWebSecurityOrigin.SubdomainSetting'
-    DisallowSubdomains = ... # type: 'QWebSecurityOrigin.SubdomainSetting'
+    class SubdomainSetting(int):
+        AllowSubdomains = ... # type: QWebSecurityOrigin.SubdomainSetting
+        DisallowSubdomains = ... # type: QWebSecurityOrigin.SubdomainSetting
```
BryceBeagle, line 275: I would assume these are missing from the containing class too.

PyQt5-stubs/QtWebKit.pyi
```diff
@@ -297,69 +297,69 @@ class QWebSecurityOrigin(sip.simplewrapper):
...
+        WebAudioEnabled = ... # type: QWebSettings.WebAttribute
+        Accelerated2dCanvasEnabled = ... # type: QWebSettings.WebAttribute
+
+    class FontFamily(int):
+        StandardFont = ... # type: QWebSettings.FontFamily
+        FixedFont = ... # type: QWebSettings.FontFamily
+        SerifFont = ... # type: QWebSettings.FontFamily
+        SansSerifFont = ... # type: QWebSettings.FontFamily
+        CursiveFont = ... # type: QWebSettings.FontFamily
+        FantasyFont = ... # type: QWebSettings.FontFamily
```
BryceBeagle: I would assume these are missing from the containing class too.

BryceBeagle (changes requested):
What an absolutely riveting PR/changelist haha

I went through everything and it looks like a couple enums were missed. I wasn't able to verify the QtWebkit stuff*, but I'd assume the pattern holds

\* Do you know which PyPI package QtWebkit is in? I couldn't locate it

PyQt5-stubs/QtWebKit.pyi
The-Compiler: Indeed - to clarify, not just `FontFamily`, but also `WebAttribute` and probably the others.

The-Compiler:
> * Do you know which PyPI package QtWebkit is in? I couldn't locate it

It's not on PyPI - given that it was [removed from Qt 5.6](https://wiki.qt.io/New_Features_in_Qt_5.6#Removed_Modules) but still (barely) [maintained independently](https://github.com/qtwebkit/qtwebkit), it's pretty much on "life support" in PyQt5. It only works when building PyQt5 from source with QtWebKit available (or when using distribution packages which do so).

It seems to be gone entirely in PyQt6 FWIW.

altendky:
Huh, I wonder what's up with stubtest missing these.  I thought it had caught such stuff.  Thanks for the review, I'll see what I can figure out.

altendky:
I guess it only catches stuff in the stubs that isn't in the runtime.  The other way could be pretty noise I suppose.  It's not like you want to hint the presence of `sys` or other misc 'not-exported' attributes in every stub etc.

BryceBeagle (approved):

PR #144, opened 2021-03-26
Add tests and fix for QTreeWidgetItem comparison (used for sorting)

bluebird75:
Error reported without the fix:
```
qtreewidgetitem.py:8: error: "__lt__" undefined in superclass
Found 1 error in 1 file (checked 1 source file)
```


BryceBeagle:
Same as other PRs: Can you add an entry to the `CHANGELOG.md`?

BryceBeagle:
Seems your revert only removed the file contents and not the file itself

bluebird75:
Indeed. This is fixed now.

PyQt5-stubs/QtWidgets.pyi
```diff
@@ -10936,6 +10936,8 @@ class QTreeWidgetItem(sip.wrapper):
     @typing.overload
     def __init__(self, other: 'QTreeWidgetItem') -> None: ...
 
+    def __lt__(self, other: QTreeWidgetItem) -> bool: ...
```
BryceBeagle: I'm surprised the type for the `other` arg doesn't need to be a forward reference... Do stubs do something different than normal Python files?

If it works, I guess we can leave it; but maybe it would be best to make it consistent with the rest of the methods on the class and wrap it in quotes.

BryceBeagle (approved):

PyQt5-stubs/QtWidgets.pyi
altendky: It's certainly plausible that quotes are not required since Mypy just parses the syntax without running it through Python.  I don't super care but I guess I'll vote for consistency for now.

```suggestion
    def __lt__(self, other: 'QTreeWidgetItem') -> bool: ...
```

altendky (changes requested):
@bluebird75, any interest in adding some more of the `__dunders__` as a big group?  I'll approve and merge without, but figured I'd 'offer'.  `:]`

Otherwise, add the quotes and we'll be good to go.  Thanks again.

PyQt5-stubs/QtWidgets.pyi
bluebird75: Forward references are supported since Python 3.7 . See https://docs.python.org/3.7/whatsnew/3.7.html#whatsnew37-pep563 . But agreed, it is better with quotes.

BryceBeagle: I think this repo is supporting Python 3.6+, plus pre-Python 3.10 you need the `from __future__ import annotations` for them to be enabled

altendky: That's about Python, not Mypy or type stubs, afaik.  Mypy reads files using https://pypi.org/project/typed-ast/, not the Python interpreter.  The only reason I know of for quotes here is consistency.  Which sometimes is pretty nice, but I can't say I actually super care about here.

https://mypy-play.net/?mypy=latest&python=3.6&gist=a3a8015378f8a40dc078f34a1a4272e6
```
class C:
    def hmm(self) -> C:
        return C()
```

```
Success: no issues found in 1 source file
```

Anyways, the quotes were added so I think we're all in agreement that this should be merged now.

altendky (approved):

altendky:
@bluebird75, thanks again for your work here.

PR #148, opened 2021-03-27
QScrollArea(widgetResizable)

### LINKED ISSUE #147: Set properties using keyword arguments in constructor
"self.scroll_area = QtWidgets.QScrollArea(widgetResizable=True)"

generates following error message:

error: Unexpected keyword argument "widgetResizable" for "QScrollArea"

altendky:
Closes #147


altendky:
Cool, this is apparently the first test to instantiate a widget...  so time to fix up CI for GUI testing.

BryceBeagle:
If you want to begin going through with the herculean effort of adding the properties to the constructor signature, I guess we might as well add [`alignment`](https://doc.qt.io/qt-5/qscrollarea.html#properties) too?

altendky:
Why stop at non-inherited ones?  `:[`  Mostly my brain went to 'someone else try to figure out how to work through all this with Phil' and 'PyQt6'.

BryceBeagle:
> Why stop at non-inherited ones?

I figured we'd just get one class out of the way, but we may want to just make this a batch change. I wonder if it's possible to write a script to determine all the Qt Properties and modify all of the stubs at once.

Edit: oh but I see what you're saying. We can't rely on inheritance for the constructor signature :sob:

PR #152, opened 2021-04-13
Implement QDialogButtonBox.StandardButton[s].__or__

erikced:
Essentially the same as #109 but for `QDialogButtonBox` which needs it as well.


altendky (approved):
My apologies for the extended delay in handling this.  Thank you for the work and the contribution.

PR #153, opened 2021-04-20
Add all qflags bases classes

bluebird75:
I started from my work on windowFlags and look for ways to make it generic. I failed at defining a generic approach with mypy, protocols and generics inheriting from int, it is really above my understanding of mypy.

But, it's easy enough to generate a file for each QFlag based class which can be used to assert both the actual implementation types and values, and which ensures that the class is being correctly checked. This is what I am doing here for two classes : WindowFlags and Alignment.

With a few more hours, I should be able to cover all qflags based class. This would remove a big hole in PyQt stubs.


PyQt5-stubs/QtCore.pyi
```diff
@@ -2755,13 +2761,19 @@ class Qt(sip.simplewrapper):
         def __invert__(self) -> 'Qt.Alignment': ...
         def __index__(self) -> int: ...
         def __int__(self) -> int: ...
+        def __or__ (self, other: typing.Union['Qt.Alignemtn', 'Qt.AlignmentFlag', int]) -> 'Qt.Alignemtn': ...
+        def __and__(self, other: typing.Union['Qt.Alignemtn', 'Qt.AlignmentFlag', int]) -> 'Qt.Alignemtn': ...
+        def __xor__(self, other: typing.Union['Qt.Alignemtn', 'Qt.AlignmentFlag', int]) -> 'Qt.Alignemtn': ...
+        def __ror__ (self, other: 'Qt.AlignmentFlag') -> 'Qt.Alignemtn': ...
+        def __rand__(self, other: 'Qt.AlignmentFlag') -> 'Qt.Alignemtn': ...
+        def __rxor__(self, other: 'Qt.AlignmentFlag') -> 'Qt.Alignemtn': ...
```
altendky:
```suggestion
        def __or__ (self, other: typing.Union['Qt.Alignment', 'Qt.AlignmentFlag', int]) -> 'Qt.Alignment': ...
        def __and__(self, other: typing.Union['Qt.Alignment', 'Qt.AlignmentFlag', int]) -> 'Qt.Alignment': ...
        def __xor__(self, other: typing.Union['Qt.Alignment', 'Qt.AlignmentFlag', int]) -> 'Qt.Alignment': ...
        def __ror__ (self, other: 'Qt.AlignmentFlag') -> 'Qt.Alignment': ...
        def __rand__(self, other: 'Qt.AlignmentFlag') -> 'Qt.Alignment': ...
        def __rxor__(self, other: 'Qt.AlignmentFlag') -> 'Qt.Alignment': ...
```

altendky (dismissed):
I'll suggest this quickly now, in case you are around and we can get the CI fixed.  I'll go back to reviewing the test code now.

tests/generate_qflags_tests.py
```diff
@@ -0,0 +1,69 @@
...
+# Content is:
+# - name of the file to generate
+# - name of the QFlags<> class
+# - name of the individual flag class (inheriting from int)
+# - one individual flag value
+# - a second individual flag value
+FLAGS_FILES = [
+    ('qflags_alignmentflags.py', 'QtCore.Qt.Alignment', 'QtCore.Qt.AlignmentFlag', 'QtCore.Qt.AlignLeft', 'QtCore.Qt.AlignRight'),
+
+]
```
altendky: There would be other changes to go with this.  I prefer to use code to document things over comments when possible.

```suggestion
@dataclasses.dataclass
class FlagsFileNames:
    generated_file: str
    flags_class: str
    flag_class: str
    first_flag_value: str
    second_flag_value: str

FLAGS_FILES = [
    FlagsFileNames(
        generated_file='qflags_alignmentflags.py',
        flags_class='QtCore.Qt.Alignment',
        flag_class='QtCore.Qt.AlignmentFlag',
        first_flag_value='QtCore.Qt.AlignLeft',
        second_flag_value='QtCore.Qt.AlignRight',
    ),
]
```

tests/generate_qflags_tests.py
```diff
@@ -0,0 +1,69 @@
...
+
+# global values for simplicity of the code
+sourcePart1, sourcePart2, sourcePart3 = [], [], []  # type: Tuple[List[str], List[str], List[str]]
```
altendky: I'm a bit of a stickler for not using globals.  I also tend pretty strongly away from `if __name__ == "__main__":` checks, but I'll skip that one here.

1) It appears that `sourcePart2` is not output anywhere.  Am I just missing it?
2) Please no globals.
3) Let's have a CI check that the generated files are in fact up to date.  Maybe run the generation script and then `git diff --exit-code` against the generated files.  Or, just generate them before running tests and don't commit them.

PyQt5-stubs/QtCore.pyi
bluebird75: I though I had fixed this one already. I think I was waiting on CI to see if it could pick it up

bluebird75:
The file where I would be interested for your review is qflags_aligmentflags.py . It uses a mostly generic approach on QFlag and QFlags based classes to check all operations on both mypy and python side.

Only two lines are specific to AlignementFlag in this file. I have work in progress to generate one similar file per QFlag based class (for the testing side) and applying automatically a fix to the pyi files to add the missing methods.

altendky (changes requested):
I personally find the generated files to be pretty terrible.  But, to be clear, that is not a judgement of your work.  Rather, mypy is a static analyzer so it forces you to do things explicitly in ways that you wouldn't do them just for 'regular' Python code.  I've had a few ideas for alternative mechanisms to get mypy to check multiple types, but they don't seem like they will actually work.

I don't know what this should look like long term, other than upstream PyQt getting this all correct, but I can go with this for now.

bluebird75:
Thanks, this is exactly the feedback I am looking for.

The use of globals was rather convenient when focusing on getting things done. I agree with you that it's better without it. Fully ok for using dataclasses, it certainly makes the thing more readable.

Just curious, how do you avoid the `if __name__ == '__main__'`  ? 

The sourcePart2 is indeed captured but not outputed anywhere. It is the part which is replaced.


For the full generation approach, I was thinking along the lines of : 
1. generate the qflag based test class
2. patch the relevant .pyi file
2. run local py.test + mypy test
3. if successful commit test file + updated pyi file, if fail, just skip the change as of now
4. move to next qflag class

One consequence of this generic approach is that there will be one test file par QFlag class, so expect quite some test files. Is that OK with you ? Or do you consider them ony as a local validation tool which does not need to go into git ?

altendky:
> The use of globals was rather convenient when focusing on getting things done. I agree with you that it's better without it. Fully ok for using dataclasses, it certainly makes the thing more readable.

Just for the record, I almost exclusively use [attrs](https://www.attrs.org/en/stable/), but didn't feel like bothering with adding the dependency here right now.  Might get swapped over later.

> Just curious, how do you avoid the `if __name__ == '__main__'` ?

In this case you just remove it.  We aren't trying to import the file so there's no need to avoid the function being called if we were to import the file.  In general, I put all modules in the package and then use [entry points](https://setuptools.readthedocs.io/en/latest/userguide/quickstart.html?highlight=entry_points#entry-points-and-automatic-script-creation) to provide the ability to run from the command line.  I rarely have directly runnable `.py` files in the projects I work on.

This is a test helper with the tests outside the package which makes it not fit this mold so well.  That's part of why I'm not pushing the matter here.

Why shouldn't we use `if __name__ == "__main__":`?  One issue is that it succeeds at allowing you to do what it is intended to.  Both import and directly run the same file.  In many cases this isn't a problem but when both happen in the same process you end up with two modules created from the single file.  It gets pretty confusing when there are two classes created from the same code that are actually different types.  I think I have an example on repl.it of this.  I can try to dig it up if you are interested.

> The sourcePart2 is indeed captured but not outputed anywhere. It is the part which is replaced.

That's probably worth at least a comment, or just not collecting that part at all.

> For the full generation approach, I was thinking along the lines of :
> 
> 1. generate the qflag based test class
> 2. patch the relevant .pyi file
> 3. run local py.test + mypy test
> 4. if successful commit test file + updated pyi file, if fail, just skip the change as of now
> 5. move to next qflag class
> 
> One consequence of this generic approach is that there will be one test file par QFlag class, so expect quite some test files. Is that OK with you ? Or do you consider them ony as a local validation tool which does not need to go into git ?

@The-Compiler, perhaps you want to chime in since you have been talking about a more automated handling of these stubs for awhile now?

I suggested not putting the generated test files in git since they are generated and could be regenerated before each test.  I could go either way though.  But if they are committed, then CI should tell us if regeneration would change them.

bluebird75:
The script is now ready. It was a bit more difficult than I expected but nothing too fancy. It is now 1000 lines, with a proper README. I will start committing some flags and see how it goes.

One issue which I am already encountering is that not all QFlags are equals. I encountered some small difference, which where caught by the strong test script. The script assumes one behavior and rejects everything else. This is fine for now, I'll look into the rejected cases another time to understand the behavior difference.

tests/generate_qflags_tests.py
bluebird75: this has been adressed now in the latest version

tests/generate_qflags_tests.py
bluebird75: This code is gone now, the super scripts replaces it.

bluebird75:
After some review, archiving a test file for each added QFlag is too much. There will be close to 100 test files just for the 10 modules listed today. When adding all other QFlags from all other modules, we should be close to 300 extra test files. This will just slow down the CI without any significant benefit. Maybe a lighter form to verify that the flag is properly supported should be the target, like checking one or two operations per flag.

altendky:
Do the tests actually take a long time?

bluebird75:
I was expecting a big impact but actually, after grouping all the qflag checks in one call, it is only 15 second impact. So very reasonable.

The branch is close to be ready for a merge. Anything else you want to see adjusted ?

bluebird75:
This is all green finally. Code is ready to be merged.

bluebird75:
Hi Kyle. Can you start reviewing that one ? It is the result of a long work so I would be thrilled to have it integrated.

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
 
 
-TESTS_DIR = os.path.dirname(__file__)
+TESTS_DIR = Path(__file__).parent
 
 
 def gen_tests():
-    for filename in os.listdir(TESTS_DIR):
-        if filename.endswith('.py') and not filename.startswith('test_'):
+    '''List of all tests files included in the directory tests'''
```
altendky: Docstrings specifically use double quotes.

```suggestion
    """List of all tests files included in the directory tests"""
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
+    for filename in TESTS_DIR.glob('*.py'):
+        if not str(filename.parts[-1]).startswith('test_'):
             yield filename
 
+def gen_abs_qflags_tests():
+    '''List of all tests included in the directory qflags'''
```
altendky:
```suggestion
    """List of all tests included in the directory qflags"""
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
+    for filename in (TESTS_DIR/'qflags').glob('test_*.py'):
+        yield filename
```
altendky:
```suggestion
    yield from TESTS_DIR.joinpath('qflags').glob('test_*.py')
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
+TESTS_DIR = Path(__file__).parent
 
 
 def gen_tests():
-    for filename in os.listdir(TESTS_DIR):
-        if filename.endswith('.py') and not filename.startswith('test_'):
+    '''List of all tests files included in the directory tests'''
+    for filename in TESTS_DIR.glob('*.py'):
+        if not str(filename.parts[-1]).startswith('test_'):
             yield filename
```
altendky: `.parts` is already strings, but the last one is `.name` anyways.

```suggestion
    for path in TESTS_DIR.glob('*.py'):
        if not path.name.startswith('test_'):
            yield path
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
+        print(stderr)
+
+    assert stdout.startswith("Success: no issues found")
+    assert not stderr
+    assert exitcode == 0
+
+
+def test_stubs_qflags() -> None:
+    """Run mypy over qflags files."""
+    stdout, stderr, exitcode = api.run([str(f) for f in gen_abs_qflags_tests()])
```
altendky:
```suggestion
    stdout, stderr, exitcode = api.run([os.fspath(path) for path in gen_abs_qflags_tests()])
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
+    '''List of all tests included in the directory qflags'''
+    for filename in (TESTS_DIR/'qflags').glob('test_*.py'):
+        yield filename
 
-@pytest.mark.parametrize('filename', list(gen_tests()))
-def test_stubs(filename):
+
+@pytest.mark.parametrize('filename',
+                         list(gen_tests()),
+                         ids=[v.parts[-1] for v in gen_tests()]
```
altendky: Could be `.name` but this avoids prescribing a single directory being handled.  Also, `.as_posix()` will give a `str` with consistent forward slashes.
```suggestion
                         ids=[v.relative_to(TESTS_DIR).as_posix() for v in gen_tests()]
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
+                         )
+def test_stubs(filename: Path) -> None:
     """Run mypy over example files."""
-    path = os.path.join(TESTS_DIR, filename)
-    stdout, stderr, exitcode = api.run([path])
+    stdout, stderr, exitcode = api.run([str(filename)])
```
altendky:
```suggestion
    stdout, stderr, exitcode = api.run([os.fspath(path)])
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
     if stdout:
         print(stdout)
+    if stderr:
+        print(stderr)
 
     assert stdout.startswith("Success: no issues found")
     assert not stderr
     assert exitcode == 0
 
+# note: no need to run explicitely pytest over qflags, because pytest finds them automatically
```
altendky:
```suggestion
# note: no need to run explicitly pytest over qflags, because pytest finds them automatically
```

tests/test_stubs.py
```diff
@@ -1,34 +1,60 @@
...
+@pytest.mark.parametrize('filename',
+                         list(gen_tests()),
+                         ids=[v.parts[-1] for v in gen_tests()]
+                         )
 def test_files(filename):
     """Run the test files to make sure they work properly."""
-    path = os.path.join(TESTS_DIR, filename)
-    with open(path, 'r') as f:
+    with open(str(filename), 'r') as f:
         code = f.read()
```
altendky: GitHub doesn't want me to suggest across the deleted line, but...

```python
    code = filename.read_text("r", encoding="utf-8")
```

altendky (approved):
Several remaining nitpicks about the use of `pathlib` (which I do very much like).  A general note I didn't point out everywhere is that I use `path` not `filename` for `pathlib.Path`s.  Feel free to merge after going through the suggestions.

For another PR, it would be good for the generation to be run in every CI run and for there to be a CI failure if the generation results in any changes, additions, or removals.

bluebird75:
Thanks for the feedback. This read_text() was unknown to me and I appreciated the other ideas.  The code was not written with top standards but to achieve a useful purpose. It probably could be improved.

It is not really feasible to run the generation on CI. First, it is very time consuming because libcst needs to parse the huge PyQt5 modules for each QFlag generation. Second, the source of the QFlag information is a manual grep over Qt sources, which requires some user modifications before being a proper source for complete generation.

But it's a good idea to run it on new Qt or PyQt releases.

PR #161, opened 2021-08-16
update requirements

altendky:


stubtest.allowlist.to_review
```diff
@@ -24,3 +24,5 @@ PyQt5.QtTest.QAbstractItemModelTester.FailureReportingMode.__new__
 PyQt5.QtWebEngineCore.QWebEngineUrlScheme.Syntax.__new__
 PyQt5.QtWebEngineWidgets.QWebEnginePage.LifecycleState.__new__
 PyQt5.QtWidgets.QActionGroup.ExclusionPolicy.__new__
+PyQt5.QtCore.pyqtProperty
+PyQt5.QtQml.QQmlListProperty
```
altendky, line 28: We just don't use these anywhere, so skipping for now.  Not sure why they show up now and not before.

PyQt5-stubs/QtWidgets.pyi
```diff
@@ -5747,7 +5747,7 @@ class QGraphicsScene(QtCore.QObject):
     def __init__(self, x: float, y: float, width: float, height: float, parent: typing.Optional[QtCore.QObject] = ...) -> None: ...
 
     def setFocusOnTouch(self, enabled: bool) -> None: ...
-    focusOnTouch: typing.ClassVar[QtCore.pyqtSignal]
+    def focusOnTouch(self) -> bool: ...
```
altendky, line 5756: Oops.  https://doc.qt.io/qt-5/qgraphicsscene.html#focusOnTouch-prop

tox.ini
```diff
@@ -3,8 +3,7 @@ envlist = py3{5,6,7,8,9}
 
 [testenv]
 deps =
-  mypy @ git+https://github.com/python/mypy@538d36481526135c44b90383663eaa177cfc32e3
-  pytest
+  -r requirements/develop.txt
   pip>=20.0
```
altendky, line 7: This is still not right, I don't think, but unrelated.  'ish.

stubtest.allowlist.to_review
bluebird75: I am working on adding pyqtProperty . Not sure why they show up either...

altendky: They should show up, I think.  The stub test is checking to make sure we have each thing documented properly in the `.pyi` files.  We don't for these.  But, I don't think we care for the sake of this dependency update PR since it isn't actually indicating any regression.  Certainly, having them in place would be good.  It seems like there will be lots of caveats for it.

bluebird75: There were probably some fixes in stubtest.py which prevented these missing methods to show up.

bluebird75:
On my branch, I went with mypy 0.800 . This is the oldest release which includes the specific commit previously referenced. I like the idea of using an old but good enough mypy release. There could be a separate entry in the CI for the latest mypy (with no version specifier) and authorization to fail . This would cover both old-mypy users and recent-mypy users.

altendky:
I'm going to go the other route here and use the latest for now.  If there's time later we can worry about trying to test against a range of mypy versions.

altendky:
I'm also going to cheat and bypass the review requirements here since I don't expect this is controversial and it has been 3+ weeks without review.

PR #162, opened 2021-08-24
Fix of all signals. Script to annotate signals automatically.

TilmanK:


rr- (approved):

bluebird75:
I love the approach and I have been using CST as well for annotating QFlags better.

I am worried though because your changes have marked incorrectly a method as a signal :

# QtWidgets.QGraphicsScene :
-    def focusOnTouch(self) -> bool: ...
+    focusOnTouch: typing.ClassVar[QtCore.pyqtSignal]

focusOnTouch is actually a property (see https://doc.qt.io/qt-5/qgraphicsscene.html) . If this one was marked incorrectly, there could be others. Could you investigate this particular case and see if it is a general problem or maybe just a PyQt5 stubs bug. While working with QFlags and CST, I have found a few buggy generated classes.

Quick reviewing the changes did not reveal anything else.

bluebird75:
I checked again. Actually, focusOnTouch is already incorrectly declared as a signal, it has nothing to do with your script. So, I'll complete the review and organize the merging.

bluebird75 (approved):

PR #164, opened 2021-08-28
Add binary operators to flag classes

rr-:
Includes a script in similar vibe to #162 

Fixes errors such as

>`error: Unsupported operand types for & ("State" and "StateFlag")`
>`note: Left operand is of type "Union[State, StateFlag]"`
>`error: Unsupported operand types for & ("KeyboardModifiers" and "KeyboardModifier")`


bluebird75:
Hi Marcin. Very nice effort. I have been running more or less the same effort in #153 . My work is a bit more extensive because I tried to cover all operations on QFlag and enum test bases. I am also using libcst for this effort.

I also discovered that not all QFlag based classes are equally implemented, some of them have special handling of the OR combination and some not.

rr-:
@bluebird75 thank you. I didn't notice your PR so sorry for the duplicated effort. Let's go with your PR. I'd appreciate if you looked at my script and checked if there's anything you like to incorporate within your PR :)

PR #166, opened 2021-09-13
Fixes annotation for QLineEdit to allow None as argument.

altendky:


bluebird75:
Try creating a QApplication with the follwing argv : `-platform minimal `

This should run on a gui-less environement.

The-Compiler:
Whoops, that was unintended!

altendky:
I would rather stick with using a full platform plugin and letting pytest-xvfb help out on Linux.

bluebird75 (approved):
Good for me if it works on all platforms of the CI, let's go this way.

altendky:
@bluebird75, you're tracking the changes more closely than I am.  Go ahead and merge this when it fits.

Issue #128, opened 2021-01-04
`QMessageBox.StandardButtons.__init__` issues


BryceBeagle:
https://github.com/stlehmann/PyQt5-stubs/blob/a3b2eeba10cbd9db4a2840a3b2a53c15d31039c2/PyQt5-stubs/QtWidgets.pyi#L6575-L6580

There are a couple issues with the `QMessageBox.StandardButtons` constructor annotations. 

Firstly, the second and third annotations are redundant; other than the args names, which I presume are both auto-generated and inaccurate. Neither are valid kwargs.

Secondly, it can also be constructed using `int` objects. As `QMessageBox.StandardButton` extends `int`, this annotation should just be replaced with using `int` objects.

```pycon
>>> QtWidgets.QMessageBox.StandardButtons(0)
<PyQt5.QtWidgets.QMessageBox.StandardButtons object at 0x7f56c949dd60>
```

TilmanK:
I guess that issue exists in various places and whoever is able to solve this receives my eternal gratitude.
```python
        progress = QProgressDialog(
            "Wait...", "Cancel", 0, 0, self
        )
        flags = progress.windowFlags()
        progress.setWindowFlags(
            flags
            & ~Qt.WindowCloseButtonHint
            & ~Qt.WindowContextHelpButtonHint
            | Qt.MSWindowsFixedSizeDialogHint
        )
```

Mypy output:

```
error: Unsupported operand types for & ("WindowFlags" and "int")
error: Argument 1 to "setWindowFlags" of "QWidget" has incompatible type "int"; expected "Union[WindowFlags, WindowType]"
```

Current possible solution: cast `flags` to int and then cast the complete statement inside `setWindowFlags` to `Qt.WindowFlags`. It's a bloodbath IMHO...

bluebird75:
I am working on a solution for all QFlags based classes

swehba:
The stubs for QMessageBox.critical(), .information(), etc. show the buttons parameter as non-optional. That causes PyCharm to generate a warning.

bluebird75:
You are right. I'll work on a simple fix.

bluebird75:
I just checked, pyqt-stubs describes a default valuefor the buttons parameter, making them effectively optional when calling critical() . I checked with PyCharm, it correctly handles this. I tested with the git version of pyqt5-stubs. Which version did you use on your side ?

bluebird75:
[...]

PR #168, opened 2021-11-09
Check optional button argument for QMessageBox

bluebird75:


bluebird75:
The job fails with something totally unrelated to QMessageBox. The CI is probably broken at the moment.

bluebird75:
PR created to check the QMessageBox problem reported in #128 but actually, there is no problem with the button argument.
Closing.

PR #169, opened 2021-11-09
Fix linux CI : Remove QtWebEngineCore.QtWebEngineCore

bluebird75:
The class is no longer present for PyQtWebEngine5.15.5  in QtWebEngineCore.pyd . The presence of this class was very probably an implementation detail which should not have made it to the stubs.


bluebird75:
This fixes the current CI.

altendky:
Just updated this and it looks to be empty with the 5.15.5 update.  Can close it unless there's something else?

bluebird75:
Sure, the goal was to fix CI. If CI is fixed, that's good!

PR #170, opened 2021-11-16
Run ci on windows

bluebird75:
Enable CI on Windows full green.

Note that: 
* I cover only 3.6 and 3.10 because intermediate versions are already covered on linux. And this is the same wheel which we are testing over and over with different Python version. Not worth the extra CPU time in my opinion
* there are some differences (bugs) between symbols available in Linux wheels and Windows wheels. This required a dedicated per-OS allowlist.

Addresses #122 for the windows part


.github/workflows/ci.yml
```diff
@@ -1,38 +1,26 @@
 name: CI
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
+      - run-ci-on-windows
```
altendky: Is the PR trigger insufficient?  Is this expected to remain and get merged?

.github/workflows/ci.yml
```diff
@@ -46,6 +34,24 @@ jobs:
...
+              name: CPython 3.6
+              tox: py36
+              action: 3.6
+          - os:
+                name: Windows
+                runs-on: windows-latest
+            python:
+                name: CPython 3.9
+                tox: py39
+                action: '3.9'
```
altendky: There are things changing in typing at boundaries in between the oldest and latest supported versions.  It's also simpler to just test everything everywhere I think, unless we have insufficient parallelism right now for all the jobs?

.github/workflows/ci.yml
```diff
@@ -55,7 +61,7 @@ jobs:
           architecture: x64
       - uses: twisted/python-info-action@v1.0.1
       - name: Install Linux Qt5 dependencies
-        if: matrix.os.matrix == 'linux'
+        if: matrix.os.name == 'Linux'
```
altendky, line 56: I have found in other Qt etc projects that it becomes worth being pretty explicit about the different uses of referencing each matrix item.

tox.ini
```diff
@@ -1,14 +1,15 @@
 [tox]
-envlist = py3{5,6,7,8,9}
+envlist = py3{6,7,8,9,10}
 
 [testenv]
 deps =
   -r requirements/develop.txt
   pip>=20.0
+passenv = RUNNER_OS
```
altendky: Might be worth being a bit more explicit rather than naming files in git based on GitHub's OS naming scheme (including case)?  Also, this requires setting this env var locally in accordance to GitHub's naming to be able to successfully run tox locally so perhaps a cli option or other mechanism would be better?

.github/workflows/ci.yml
```diff
@@ -1,38 +1,26 @@
 name: CI
 
 on:
+  workflow_dispatch:
```
altendky: What does this achieve?  I haven't worked with it before.

.github/workflows/ci.yml
bluebird75: No, it's a leftover. I needed it to trigger the CI on my own repo, for this branch. It needs to go now.

.github/workflows/ci.yml
bluebird75: Ok. I don't like using too much power in general, hence the attempt to be reasonably minimalistic.

.github/workflows/ci.yml
bluebird75: Sure. Maybe that one needs to go away.

tox.ini
bluebird75: Tox is just so long to run that I use it only for the CI. But indedd, on a local computer, this beomes an issue. I am quite unfamiliar with tox, what would be a good aproach for this ?

.github/workflows/ci.yml
bluebird75: It allows me to manually trigger a CI build on my repo, where not every branch is built by default. Not really needed for this repo here though

.github/workflows/ci.yml
altendky: That's good.  Don't forget the relatively insane cost of an undetected bug.  `:]`

tox.ini
altendky: I guess "the tox way" would actually be an environment factor.  Something like below.  Then running tox you would `tox -v -e py39-windows`, or in the CI config something like `tox -v -e ${{ matrix.python.tox }}-${{ matrix.os.tox }}`.

```python
[testenv]
deps =
  -r requirements/develop.txt
  pip>=20.0
setenv:
  windows: OS_FILE_MARKER = windows
  linux: OS_FILE_MARKER = linux
commands = 
  pip install PyQt5==5.15.* PyQt3D==5.15.* PyQtChart==5.15.*  PyQtDataVisualization==5.15.* PyQtNetworkAuth==5.15.* PyQtPurchasing==5.15.* PyQtWebEngine==5.15.*
  mypy --show-error-codes -p PyQt5-stubs
  stubtest --allowlist {toxinidir}/stubtest.allowlist --allowlist {toxinidir}/stubtest.allowlist.to_review --allowlist {toxinidir}/stubtest.allowlist.{env:OS_FILE_MARKER} PyQt5
  pytest --capture=no --verbose {posargs}
```

.github/workflows/ci.yml
```diff
@@ -1,38 +1,26 @@
...
+      - run-ci-on-windows
     tags:
       - v*
   pull_request:
     branches:
       - "*"
-  schedule:
-    # Daily at 06:14.  This is just an arbitrary time to stagger
-    # relative to other daily builds in any given organization.
-    - cron: '14 6 * * *'
```
altendky: Another adjustment for CI in your repo?  I generally encourage people to use the "real" CI, personally.  More consistency, less noise.

.github/workflows/ci.yml
```diff
@@ -46,6 +37,46 @@ jobs:
...
+              tox: py39
+              action: 3.9
+# Windows 10 not working yet, because there are no wheels yet for it.
+#         - os:
+#               name: Windows
+#               runs-on: windows-latest
+#           python:
+#               name: CPython 3.10
+#               tox: py310
+#               action: '3.10'
```
altendky: Maybe just add Windows to the matrix (uncomment in this case) like in https://github.com/python-qt-tools/PyQt5-stubs/pull/123/files?  Then add 3.10 separately and you can use an exclude if needed rather than including all the other cases.

(and don't delete the commented out macOS config since it is its own thing)

PyQt5-stubs/QtCore.pyi
```diff
@@ -12297,6 +12297,67 @@ class QSysInfo(sip.simplewrapper):
 
     WordSize = ...  # type: QSysInfo.Sizes
 
+    class WinVersion(int):
```
altendky, line 12299: I guess this should be present if we built the upstream on Windows?  Which raises an annoying question about how to handle upstream updates on multiple platforms.  `:[`

bluebird75: I guess so. But anyway, teststub is very good at catching this. So, you already have an answer : we need to make sure to have CI run on all platforms.

In this case, these are deprecated functions, so no worry.

tox.ini
bluebird75: done as described.

bluebird75:
I think we are good now to merge.

tox.ini
```diff
@@ -1,16 +1,19 @@
 [tox]
-envlist = py3{5,6,7,8,9}
+envlist = py3{6,7,8,9,10}-{windows,linux}
```
altendky, line 2: This is not a concern to be addressed in this PR, but since this change points it out again...

https://github.com/python-qt-tools/PyQt5-stubs/blob/78f4dc175cd05357421e05a27ef0733f0bcd1736/setup.py#L50

We're in a bit of a sticky situation where we only test against, and generally speaking are only "correct" for the latest version of PyQt5.  That doesn' support Python 3.5 so we don't test on 3.5.  But, this package might still be useful on versions less than 3.6.  Maybe?  Anyways, it's something to think about.  When do we bump the above line to require 3.6?  I'll also note that 3.6 is at least scheduled to go EOL this month.

.github/workflows/ci.yml
```diff
@@ -46,6 +41,15 @@ jobs:
...
             action: 3.9
+          - name: CPython 3.10
+            tox: py310
+            action: '3.10'
+        exclude:
+# Windows 10 not working yet, because there are no wheels yet for it.
+         - os:
+               name: Windows
+           python:
+               name: CPython 3.10
```
altendky: Looks ok over at https://github.com/python-qt-tools/PyQt5-stubs/runs/4411387242?check_suite_focus=true from https://github.com/python-qt-tools/PyQt5-stubs/pull/180.  Note that the wheels are abi3 https://pypi.org/project/PyQt5/5.15.6/#files so they should be compatible with any new version of Python3.  Though, that was introduced in 5.15.5 which we only recently got updated past.

```suggestion
```

altendky (changes requested):
I'll approve when 3.10 is included on Windows, presuming my check for that is accurate.  Thanks!

tox.ini
bluebird75: Indeed, it is a good idea to keep 3.5 running. At my workplace, we recently settled on using 3.5 instead of using a mix of 3.1 and 3.5 . Old python versions still have a long life in front of them.

.github/workflows/ci.yml
bluebird75: Good that it is working. This was after my post to PyQt5 mailing list that Phil released new versions.

altendky (approved):

PR #171, opened 2021-11-20
More qflags

bluebird75:
Add QFlag related methods to all modules within PyQt5-stubs. This is to be merged after #153 .


tests/qflags/grep-results/qt3d-qflag-grep-result.txt
```diff
@@ -0,0 +1,8 @@
+src\animation\frontend\qanimationcallback.h:    Q_DECLARE_FLAGS(Flags, Flag)
```
altendky, line 1: These are supposed to be committed?

tests/qflags/qflags_modules_analysis.json
```diff
@@ -0,0 +1,1920 @@
+{
```
altendky, line 1: Or these?

tests/test_stubs.py
```diff
@@ -1,34 +1,59 @@
...
 
 
 def gen_tests():
-    for filename in os.listdir(TESTS_DIR):
-        if filename.endswith('.py') and not filename.startswith('test_'):
-            yield filename
+    """List of all tests files included in the directory tests"""
+    for path in TESTS_DIR.glob('*.py'):
+        if not path.name.startswith('test_'):
+            yield path
```
altendky:
```suggestion
    yield from TESTS_DIR.glob('test_*.py')
```

tests/test_stubs.py
```diff
@@ -1,34 +1,59 @@
...
     assert not stderr
     assert exitcode == 0
 
+# note: no need to run explicitly pytest over qflags, because pytest finds them automatically
 
-@pytest.mark.parametrize('filename', list(gen_tests()))
-def test_files(filename):
+@pytest.mark.parametrize('filepath',
+                         list(gen_tests()),
+                         ids=[v.parts[-1] for v in gen_tests()]
```
altendky:
```suggestion
                         ids=[v.name for v in gen_tests()]
```

altendky (commented):
There are a lot of new files.  Perhaps they should be grouped by the module they are for?

tests/qflags/grep-results/qt3d-qflag-grep-result.txt
bluebird75: Yes, this gives the reproducibility. Without this, it looks a bit like black magic to identify what is a potential QFlag. With it, it is possible to run (manually) the full process and reach the same result.

bluebird75:
Grouping the files could make sense althought it means spending again some time on the generation scripts. If that's OK for you, I would like to commit them asis and update them by grouping them later in 2022.

altendky (approved):
Sure.

tests/test_stubs.py
bluebird75: the suggestion changes the code. We want the path not starting with test_

altendky: Uh, right...  Sorry.

PR #173, opened 2021-11-28
drop requirements files and pin to the supported 5.15.3

altendky:


bluebird75:
This looks good to me. The CI will no longer break overnight.

PR #181, opened 2021-12-05
Make some Windows specific hints platform sensitive

altendky:


bluebird75:
interesting. I tried the same approach using platform.system() == "Windows" but it would not pass through teststub.

bluebird75 (approved):
Looks good to me

altendky:
Yeah, mypy is only so aware...  it also sometimes matters where the checks are located.

PR #182, opened 2021-12-15
Add tests for qsize

bluebird75:
Add tests around QSize operations.


bluebird75:
Oups, bad way to propose it

PR #185, opened 2022-01-02
Support pyqt properties

bluebird75:


bluebird75:
I was not able to get a full support. Support of python property seems to be hardcoded into mypy and may not be described successfully using stub syntax. In any case, there is enough support here for a merge.

.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_dispatch:
```
altendky:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -80,3 +81,6 @@ jobs:
         shell: python
         run: |
           import this
+          
+          
+
```
altendky:
```suggestion
```

altendky (commented):
This seems to be missing the core aspect of properties being descriptors that exist as class attributes.  `.__get__()` etc.

Here's the small example I tried to see what is handled.

```python
import typing

from PyQt5 import QtCore


class C(QtCore.QObject):
    pc: QtCore.pyqtProperty = QtCore.pyqtProperty(type=int)
    pccv: typing.ClassVar[QtCore.pyqtProperty] = QtCore.pyqtProperty(type=int)

    @QtCore.pyqtProperty(type=int)
    def pd(self) -> int:
        return 5

    @pd.setter
    def pd_setter(self, value: int) -> None:
        pass

    def pcf_setter(self, value: int) -> None:
        pass

    pcf = QtCore.pyqtProperty(type=int, fset=pcf_setter)
    pcfc: typing.ClassVar[QtCore.pyqtProperty] = QtCore.pyqtProperty(type=int, fset=pcf_setter)


def main() -> None:
    application = QtCore.QCoreApplication([])
    c = C()
    c.pc = 3
    c.pccv = 3

    # just exploring the handling of the not-None case
    assert c.pccv.fget is not None
    x: int = c.pccv.fget()
    y: str = c.pccv.fget()

    c.pcf = 5
    c.pcf = ""
    # just exploring the handling of the not-None case
    assert c.pcf.fset is not None
    c.pcf.fset(7)
    c.pcf.fset("")

    c.pcfc = 5
    c.pcfc = ""
    # just exploring the handling of the not-None case
    assert c.pcfc.fset is not None
    c.pcfc.fset(7)
    c.pcfc.fset("")


main()
```

```console
$ venv/bin/mypy --strict x.py
x.py:10: error: Value of type variable "TPropGetter" of "__call__" of "pyqtProperty" cannot be "Callable[[C], int]"
x.py:14: error: Value of type variable "TPropSetter" of "setter" of "pyqtProperty" cannot be "Callable[[C, int], None]"
x.py:28: error: Incompatible types in assignment (expression has type "int", variable has type "pyqtProperty")
x.py:29: error: Cannot assign to class variable "pccv" via instance
x.py:29: error: Incompatible types in assignment (expression has type "int", variable has type "pyqtProperty")
x.py:36: error: Incompatible types in assignment (expression has type "int", variable has type "pyqtProperty")
x.py:37: error: Incompatible types in assignment (expression has type "str", variable has type "pyqtProperty")
x.py:43: error: Cannot assign to class variable "pcfc" via instance
x.py:43: error: Incompatible types in assignment (expression has type "int", variable has type "pyqtProperty")
x.py:44: error: Cannot assign to class variable "pcfc" via instance
x.py:44: error: Incompatible types in assignment (expression has type "str", variable has type "pyqtProperty")
Found 11 errors in 1 file (checked 1 source file)
```

bluebird75:
I have not succeeded in describing pyqtProperty to support all its capabilities. The current status is an improvement over the nothing we have right now.

I would like to go with merging the current pull request as it is and prepare a PyQt5-stubs release. And keep the improvements for the next release.

PR #188, opened 2022-01-03
Qscrollarea widgetresizeable

bluebird75:
Finish the job started in #147 with simpler gui-less approach


altendky:
Is the plan to allow PRs that piecemeal add the dozens of parameters that PyQt adds to the constructor for the properties?

bluebird75:
There is a grand plan for adding all properties and signals to the constructor of all QObject derived classes. But this is a huge plan. In the meantime, a user suggested this valid improvement, I'm all for taking it in. More examples are always a good thing in any case.

altendky (changes requested):
I think the application creation can be dropped once https://github.com/python-qt-tools/PyQt5-stubs/pull/166 goes in, then this looks good.

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * [#183](https://github.com/python-qt-tools/PyQt5-stubs/pull/183) Add missing operations on QSize
 * [#189](https://github.com/python-qt-tools/PyQt5-stubs/pull/189) Fix QListWidget, QTreeWidget and QTableWidget so that their respective items are now optional in many places when used as argument or return value
 * [#165](https://github.com/python-qt-tools/PyQt5-stubs/pull/165) allow `None` as argument for `QLineEdit.setText()`
+* [#148](https://github.com/python-qt-tools/PyQt5-stubs/pull/148) add `widgetResizable` parameter to `QScrollArea.__init__()` and
+    `objectName` parameter to `QGroupBox.__init__()` .
```
altendky: I would tend to make all text, at least outside of Python code, use semantic newlines.

```suggestion
* [#148](https://github.com/python-qt-tools/PyQt5-stubs/pull/148) add `widgetResizable` parameter to `QScrollArea.__init__()` and `objectName` parameter to `QGroupBox.__init__()`.
```

altendky (approved):

Issue #151, opened 2021-04-13
Problem with Qt methods that return pointers


TilmanK:
I just stumbled over an issue, that I was well aware of, but that wasn't reported by MyPy since it's incorrectly annotated in the stubs. Nevertheless, I think this is more of a stub-wide problem. Here's a minimal example:

```import sys

from PyQt5.QtWidgets import QTableWidget, QApplication

app = QApplication(sys.argv)

# create a QTableWidget with one row and one column
table_widget = QTableWidget()
table_widget.setRowCount(1)
table_widget.setColumnCount(1)

# Get the item in Cell(0,0)
item = table_widget.item(0, 0)

# will print: None
print(str(item))

sys.exit(app.exec())
```

`QTableWidget.item()` returns a `QTableWidgetItem` per annotation. Nevertheless, the C++ method returns a pointer which can (and is in the shown example) a null pointer or None in Python.

How can we take care of this? In QTableWidget alone there are about ten methods that return a pointer. Shouldn't every single one of them have an `Optional ` around their return type? My C++ days are long gone, so I'm not sure: can all of those methods return a null pointer?

The-Compiler:
> so I'm not sure: can all of those methods return a null pointer?

In theory, anything can return a null pointer - with raw pointers like used by Qt, there's no distinction between `SomeType` and `Optional[SomeType]`. However, since Qt uses pointers everywhere, annotating everything as `Optional` would be cumbersome. Many of those things probably are never `None` in practice.

Thus, this is something which needs to be fixed on a case-by-case basis based on Qt's documentation - for [QTableWidget::item](https://doc.qt.io/qt-5/qtablewidget.html#item) that indeed says (emphasis mine):

> Returns the item for the given row and column if one has been set; **otherwise returns `nullptr`**.

Similarly, `horizontalHeaderItem`, `itemAt`, and `mimeData` (only if passing an empty list?) are documented that way. I suspect `verticalHeaderItem` and the `take...` variants do so as well, but to be certain, one would've to check the code.

TilmanK:
@The-Compiler thanks for the explanation, yeah, that question was kind of stupid. Of course those functions return what was intended when developing them.

Ok. Maybe I'll find some time to fix that stuff at least for common models, views and the widget classes combining them since I tend to use them quite often.

bluebird75:
@TilmanK  any news on this ?

Issue #69, opened 2020-09-21
Plans for PyQt6?


BryceBeagle:
PyQt6 snapshots [just came out.](https://www.riverbankcomputing.com/pipermail/pyqt/2020-September/043201.html) Currently they're nothing more than PyQt5 with some deprecated features removed. But it begs the question: will `PyQt5-stubs` support PyQt6?

altendky:
I haven't looked hard at 6 but it sounds like there are the expected backwards incompatible changes you would expect from a major version, even when wrapping Qt5.  (the enum stuff?) I imagine the answer being 'no it does not' but that it would be nice to.  I have no idea on the timeline for Qt6 but i tend to like moving forward sooner than later so I will likely be on-board for working on the support.  But for now I think getting cleaned up with 5 and seeing if we can work with Phil to get proper fixes upstream would be my priority.  Then there's less for us to deal with (theoretically nothing, but...).

BryceBeagle:
PyQt6 was just officially released: https://www.riverbankcomputing.com/pipermail/pyqt/2021-January/043485.html

Also of interest is that the OAuth stuff is now moved to its own package, `PyQt6-NetworkAuth`. The stubs would probably need to account for this

TilmanK:
Guys, I just migrated one of my projects to PyQt6. No stubs. It's horrible ;)

Do you have any plans so far? Getting the PyQt5 stubs at least to work isn't that hard (tried it here: https://github.com/TilmanK/PyQt5-stubs), nevertheless, I think that you generated them somehow automatically in the past?

I know that they changed some arguments (`pos` is now `position` at some places), you need to use Enums now instead of class properties (i.e. `QEvent.Type.Close` instead of `QEvent.Close`) - so the PyQt5 hints aren't that useful either. Additionally, to what @BryceBeagle already wrote above.

TilmanK:
I dug a bit and found that the generated stubs come with PyQt6 itself. So this is basically just a lot of manual work to merge them together, right?

Before I waste my weekend on this: Have you done anything yet for PyQt6? How do you plan on integrating it into this repo (do you even? - doesn't seem like the best thing to do)...

altendky:
[...]

PR #189, opened 2022-01-04
Fix annotations for QTableWidget #151

TilmanK:
Ok, so here's the PR. I've tested every method and indeed every one of them either returns a `None` when called with invalid values or accepts `None` as argument, which is convenient regarding use-cases like this:

```py
row = table_widget.row(table_widget.itemAt(QPoint(42, 42)))
```

I'll update the PR later today with fixes for `QListWidget` and `QTreeWidget` as well.


TilmanK:
@bluebird75 Go ahead, it's complete. I've confirmed in simple examples that the changes are correct - you can really build complex widget structures just by passing None around ;)

bluebird75 (approved):

bluebird75:
Nice.

Unrelated but I want to get started on PyQt6-stubs. Can you rejoin the thread on #69 ? I see that you have a repo with some code for it.

mkrieger1:
I believe `QTableWidget.cellWidget` should return `Optional[QWidget]`, not `List[QWidget]`. Can someone confirm that `List[QWidget]` was a mistake, or am I wrong?

(I already wrote an inline comment [here](https://github.com/python-qt-tools/PyQt5-stubs/commit/cd0c74401d9d4fb50761b9849e7a229d6470c4d7), but I'm not sure if this triggered a notification for anybody involved.)

altendky:
https://doc.qt.io/qt-5/qtablewidget.html#cellWidget
> `[QWidget](https://doc.qt.io/qt-5/qwidget.html#QWidget) *QTableWidget::cellWidget(int row, int column) const`

Looks like it shouldn't be a `List`, yeah.  It doesn't clearly state that the pointer can be `NULL`, but I'd guess it can be.  Maybe you could explore at runtime to double check?  And provide a pr?

mkrieger1:
Created #198.

PR #195, opened 2022-04-26
Add more tests

bluebird75:


bluebird75:
I reviewed all PR, issues and commit since the beginning of the projects. I tried to add tests for most things which were actually changed during this lifetime, with the exception of signals and enum which are handled by automation scripts.

The idea for me is to capture all user contributions into the test suite, which can serve as a basis for other pyqt-stubs projects.

bluebird75:
Ready to be merged

PyQt5-stubs/QtWidgets.pyi
```diff
@@ -10572,11 +10578,11 @@ class QTableWidgetItem(sip.wrapper):
...
     def __eq__(self, other: object) -> bool: ...
-    def __ge__(self, other: object) -> bool: ...
-    def __gt__(self, other: object) -> bool: ...
-    def __le__(self, other: object) -> bool: ...
-    def __lt__(self, other: object) -> bool: ...
     def __ne__(self, other: object) -> bool: ...
+    def __gt__(self, other: 'QTableWidgetItem') -> bool: ...
+    def __le__(self, other: 'QTableWidgetItem') -> bool: ...
+    def __lt__(self, other: 'QTableWidgetItem') -> bool: ...
+    def __ge__(self, other: 'QTableWidgetItem') -> bool: ...
```
altendky: Unless there's a reason for order to change here and differ from the `QListWidgetItem` case above?

```suggestion
    def __ge__(self, other: 'QTableWidgetItem') -> bool: ...
    def __gt__(self, other: 'QTableWidgetItem') -> bool: ...
    def __le__(self, other: 'QTableWidgetItem') -> bool: ...
    def __lt__(self, other: 'QTableWidgetItem') -> bool: ...
```

PyQt5-stubs/QtWidgets.pyi
```diff
@@ -11316,7 +11322,12 @@ class QTreeWidgetItem(sip.wrapper):
     @typing.overload
     def __init__(self, other: 'QTreeWidgetItem') -> None: ...
 
+    def __eq__(self, other: object) -> bool: ...
+    def __ne__(self, other: object) -> bool: ...
     def __lt__(self, other: 'QTreeWidgetItem') -> bool: ...
+    def __gt__(self, other: 'QTreeWidgetItem') -> bool: ...
+    def __le__(self, other: 'QTreeWidgetItem') -> bool: ...
+    def __ge__(self, other: 'QTreeWidgetItem') -> bool: ...
```
altendky: As above.

```suggestion
    def __ge__(self, other: 'QTreeWidgetItem') -> bool: ...
    def __gt__(self, other: 'QTreeWidgetItem') -> bool: ...
    def __le__(self, other: 'QTreeWidgetItem') -> bool: ...
    def __lt__(self, other: 'QTreeWidgetItem') -> bool: ...
```

tests/pyqtsignal.py
```diff
@@ -5,11 +5,25 @@
 class Class(QtCore.QObject):
     signal = QtCore.pyqtSignal([str])  # type: typing.ClassVar[QtCore.pyqtSignal]
 
+    def __init__(self) -> None:
+        super().__init__()
```
altendky, line 9: Does defining this have an affect vs. letting the `QObject.__init__()` handle it?

tests/pyqtslot.py
```diff
@@ -8,9 +10,84 @@ def func_none(s: str) -> None:
 def func_int(s: str) -> int:
     return 42
 
-@pyqtSlot(str, result='int')
+@pyqtSlot(str, result='int')    # this means int here
 def func_str(s: str) -> str:
```
altendky, line 14: Is this testing incorrect usage?  I would expect that `result='int'` would correspond with `-> int:`.  Though I've never really dealt much with slot return values.

tests/pyqtslot.py
```diff
@@ -8,9 +10,84 @@ def func_none(s: str) -> None:
...
     return '42'
 
-func_none("test")
-x = func_int("test")  # type: int
+@pyqtSlot(str, name='toto', revision=33, result=int)
+def func_toto(s: str) -> int:
+    return 33
+
+n = None
+n = func_none("test")
```
altendky, line 22: Is this better than the existing explicit hinting of the variable type?

tests/qtest.py
```diff
@@ -0,0 +1,4 @@
+from PyQt5.QtTest import QTest
+
+# check that this is a static method
+QTest.qWait(33)
```
altendky: Trailing EOL.

tests/pyqtsignal.py
```diff
@@ -5,11 +5,25 @@
...
 instance.signal.disconnect
 instance.signal[str].emit
+
+# use some of them
+connection = QtCore.QMetaObject.Connection()
+
+connection = instance.signal.connect(instance.my_slot)
+instance.signal.disconnect()
+connection = instance.signal.connect(instance.my_slot)
+instance.signal.disconnect(connection)
```
altendky: Trailing EOL.

PyQt5-stubs/QtWidgets.pyi
bluebird75: Not exactly, I just moved the __ne__ closer to __eq__ . The other changes might have slipped in.

tests/pyqtsignal.py
bluebird75: I can not reproduce right now, but I remember going crazy at first when that line was not present. Mypy would refuse to consider instance.signal as a typed variable and would output "Any" when using reveal_type(). The only way to fix that problem was to add this explicit initialization.

The problem is that assigning incorrectly Any to a type is not detected by the current test. I just checked, removing the __init__ line does not trigger the behavior I saw, but I remaing wary of it. So, I prefer to keep it.

tests/pyqtslot.py
bluebird75: I find it more reliable and more explicit. So it is better but in a very subjective way. In theory, mypy can catch the absence of Optional accurately in both cases.

tests/pyqtslot.py
bluebird75: This was introduced long before I joined the project. I agree it's hard to interpret correctly, that's certainly what the initial author wanted to test.

bluebird75:
@altendky do you want to review the changes after your review ? I believe this is ready to be merged.

altendky (approved):
Thanks for the continued work on this.

PR #197, opened 2022-06-20
Remove PyQt5.sip.array from allowlist

altendky:


altendky:
Given the failure on `master`, presumably this will break 3.6 though I haven't identified why quite yet.

altendky:
https://www.diffchecker.com/QDxLmKiR shows `PyQt5-sip==12.9.1` vs. `PyQt5-sip==12.11.0` so I guess that's the issue.

altendky:
I don't necessarily like this, but it's an option and confirms the problem.

bluebird75 (approved):
Approved. I am just concerned generally of how many specific allowlist we have to manage

bluebird75:
The unbreakage seems to come from PyQt5-sip moving from 12.10.1 to 12.11.0 for python > 3.6 .

PR #199, opened 2022-11-17
add default for `owner` parameter of `pyqtSignal.__get__()`

altendky:
https://github.com/python-qt-tools/PyQt5-stubs/actions/runs/3485949394/jobs/5831945350#step:7:111
> `error: PyQt5.QtCore.pyqtSignal.__get__ is inconsistent, runtime argument "owner" has a default value but stub argument does not`


The-Compiler (changes requested):
Does it really need to be conditional? Looking at the commit history, it looks like the `Optional` is what made it work. From reading [the Python 3.8 reference](https://docs.python.org/3.9/reference/datamodel.html#object.__get__) and [even the original 2001 PEP](https://peps.python.org/pep-0252/#specification-of-the-attribute-descriptor-api) (*`__get__()`: a function callable with one or two arguments*), it looks like this would be the correct annotation even for older Python versions.

altendky:
Adding `Optional` with a `...` default fails on everything except 3.10.
https://github.com/python-qt-tools/PyQt5-stubs/pull/200
https://github.com/python-qt-tools/PyQt5-stubs/actions/runs/3508337017/jobs/5876674894#step:7:111
```
error: PyQt5.QtCore.pyqtSignal.__get__ is inconsistent, stub argument "owner" has a default value but runtime argument does not
Stub: at line 58
Overload(def (self: PyQt5.QtCore.pyqtSignal, instance: None, owner: Union[Type[PyQt5.QtCore.QObject], None] =) -> PyQt5.QtCore.pyqtSignal, def (self: PyQt5.QtCore.pyqtSignal, instance: PyQt5.QtCore.QObject, owner: Union[Type[PyQt5.QtCore.QObject], None] =) -> PyQt5.QtCore.pyqtBoundSignal)
Inferred signature: def (self: PyQt5.QtCore.pyqtSignal, instance: Union[None, PyQt5.QtCore.QObject], owner: Union[Type[PyQt5.QtCore.QObject], None] = ...)
Runtime:
def (self, instance, owner, /)
```

Adding just `Optional` fails only on 3.10.
https://github.com/python-qt-tools/PyQt5-stubs/pull/201
https://github.com/python-qt-tools/PyQt5-stubs/actions/runs/3508353763/jobs/5876702368#step:7:111
```
error: PyQt5.QtCore.pyqtSignal.__get__ is inconsistent, runtime argument "owner" has a default value but stub argument does not
Stub: at line 58
Overload(def (self: PyQt5.QtCore.pyqtSignal, instance: None, owner: Union[Type[PyQt5.QtCore.QObject], None]) -> PyQt5.QtCore.pyqtSignal, def (self: PyQt5.QtCore.pyqtSignal, instance: PyQt5.QtCore.QObject, owner: Union[Type[PyQt5.QtCore.QObject], None]) -> PyQt5.QtCore.pyqtBoundSignal)
Inferred signature: def (self: PyQt5.QtCore.pyqtSignal, instance: Union[None, PyQt5.QtCore.QObject], owner: Union[Type[PyQt5.QtCore.QObject], None])
Runtime:
def (self, instance, owner=None, /)
```

And yeah, this does feel odd that the Python version is changing the same compiled version of PyQt's presence or lack of a parameter default...  We are on mypy 0.930, so there's that.  And 0.991 segfaults on the stubtest run so...

https://github.com/python-qt-tools/PyQt5-stubs/actions/runs/3508399647/jobs/5876778073#step:7:111
```
ERROR: InvocationError for command /home/runner/work/PyQt5-stubs/PyQt5-stubs/.tox/py310-linux/bin/stubtest --allowlist stubtest.allowlist --allowlist stubtest.allowlist.to_review --allowlist stubtest.allowlist.linux PyQt5 (exited with code -11 (SIGSEGV)) (exited with code -11)
```

altendky:
We're also on PyQt 5.15.6 not 5.15.7.

The-Compiler (approved):
Weird! But ok, let's do this, to make stubtest happy.

PR #200, opened 2022-11-20
Fixup ci example failure - optional with default

altendky:


bluebird75:
I am catching up late. This one can be closed, right ? You already fixed everything in #199

PR #201, opened 2022-11-20
Fixup ci example failure 2

altendky:


bluebird75:
I am confused about this pull request. Isn't it also the same as #199 and #200 . Was it for expirmenting with the different solutions ?

altendky:
Yes, it was a demo linked from the explanation in https://github.com/python-qt-tools/PyQt5-stubs/pull/199#issuecomment-1321148787.

PR #202, opened 2022-11-20
mypy==0.991

altendky:


bluebird75:
This version of mypy is not released for Python 3.6 . So we need to add a conditional, one way or the other...

Apart from that, we can also search which was the first version that was broken by our stubs. I don't have much time right now for this, though.

altendky:
Locally here in Linux, 0.940 is the first version to `SIGSEGV`.  With 0.960 through 0.991 I get a few messages first but then they also `SIGSEGV`.  https://github.com/python/mypy/commit/a9c62c5f82f34a923b8117a5394983aefce37b63 acts similar to 0.991.  There are messages followed by a `SIGSEGV`, though there's an additional message.  See https://github.com/python/mypy/issues/14196.

setup.py
```diff
@@ -51,7 +51,12 @@ def find_version(*file_paths):
...
     packages=["PyQt5-stubs"],
     extras_require={
-        "dev": ["mypy==0.991", "pytest", "pytest-xvfb"],
+        "dev": [
+            "mypy==0.991; python_version >= '3.7'",
+            "mypy==0.930; python_version < '3.7'",
+            "pytest",
+            "pytest-xvfb",
+        ],
     },
```
bluebird75, line 67: I did not know you can do things like this. Nice!

bluebird75:
I am convinced that we should have a very stable mypy for our tests, and one CI job with failure allowed for testing latest mypy.

Until this is fixed, 0.930 is our good version then.

setup.py
altendky: See the `env_var` list at https://peps.python.org/pep-0508/.

```
env_var       = ('python_version' | 'python_full_version' |
                 'os_name' | 'sys_platform' | 'platform_release' |
                 'platform_system' | 'platform_version' |
                 'platform_machine' | 'platform_python_implementation' |
                 'implementation_name' | 'implementation_version' |
                 'extra' # ONLY when defined by a containing layer
                 )
```

Though it appears this boundary may belong closer to 0.981 https://github.com/python/mypy/commit/dc118e293203863ab1007699b2cecf0f26ddfa22.

altendky:
Dependabot style automatic PRs are an alternative to consider instead of having 'latest version' jobs.  But, there are upsides to all of various approaches to this.

Reported the segfault upstream.  https://www.riverbankcomputing.com/pipermail/pyqt/2022-November/045068.html  I'll take a look at a stubtest wrapper that monkey patches to disable the class finality test.

altendky:
segfaults gone...  now just 2853 regular stubtest errors.  Hopefully my comments around the dependency versions in `setup.py` are sufficient to have a clue about what we're stuck on here.

bluebird75:
At least it passes with Python 3.6

altendky:
Note that there's mypy 1.0 now as well.  Depending, it may be better to work this intermediate step, or may be better to just get to 1.0 directly.

PR #206, opened 2022-11-23
update actions/checkout and actions/setup-python to avoid warnings

altendky:
https://github.com/python-qt-tools/PyQt5-stubs/actions/runs/3529777349
> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-python@v2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,9 +52,9 @@ jobs:
             tox: py310
             action: '3.10'
     steps:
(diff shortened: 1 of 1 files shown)


bluebird75 (approved):
Looks good to me.

PR #212, opened 2023-06-29
Make QWebEngineView.setPage() parameter optional

sanjacob:
This change allows the page passed to setPage() to be None, which is a valid technique and often needed.


bluebird75:
Thanks for the proposal (and sorry for the long delay).

